PR #60, opened 2022-11-02
curriculum: Don't print error if RE[] tag is missing

tomtseng:
## Description

* When pass hardening on, we more frequently see games that hit the move limit of 1600 moves
* Games that hit the move limit don't have a RE[] tag
* We shouldn't log an error when we don't see an RE[] tag or else we may see a bunch of logged errors on every iteration of `curriculum.py`

## Testing

Ran `curriculum.py` on a batch of existing SGFs and no `RE[]` errors appeared


AdamGleave (approved):
LGTM

PR #61, opened 2022-11-02
searchresults: Suppress playing in pass-alive-territory if using pass-alive hardening

### LINKED ISSUE #58: Fix loophole in `avoid-pass-alive-territory`
From @tomtseng:
> Here’s a board position where white is the hardened victim. We suppress passing for the hardened victim because there is at least one legal move not in its pass-alive territory. In this case there is only one such move (and it’s not a good move), marked with a black triangle in the diagram. Now the victim picks moves among all legal non-passing moves, including moves in its pass-alive territory. Over the next 60 moves, the victim fills in the eyes in its pass-alive territory and loses nearly all of its stones 

Screen Shot 2022-10-25 at 15 25 31

> maybe the answer is just to not only suppress passing but to also suppress all moves in the victim’s pass-alive territory. Then the victim would be forced to play that one move marked with the black triangle. That’s a bad move because it loses 3 stones, but katago’s scoring analysis marks those 3 squares as being controlled by black anyway

tomtseng:
### Context

Fixes #58 

`passingBehavior = avoid-pass-alive-territory` hardening makes the bot not pass if there are any legal moves outside the bot's pass-alive territory. This can lead to bad behavior where the bot in which the bot makes the obviously bad move of filling in eyes in its own pass-alive territory and eventually losing its pass-alive territory. Probably what's happening here is that the bot's policy NN says the best move by far is to pass, but then we suppress passing and the bot is left to select among a bunch of bad low-confidence moves.

### Changes

* Patch `passingBehavior = avoid-pass-alive-territory` so that playing within one's own pass-alive territory is always suppressed

Note:
* pass-alive territory is not suppressed during MCTS rollouts. If all MCTS rollouts only explore suppressed moves, then we just use the policy net directly to pick moves
  * (Maybe I should suppress moves within MCTS search too? Otherwise a lot of visits could be wasted on moves that ultimately are suppressed. Boosting the visit count may be enough to compensate for wasted visits. It might be interesting to gather the mean/median of "effective" visits (i.e., number of visits minus all visits that hit suppressed moves) in a set of games)

## Testing

Enabled `avoid-pass-alive-territory` for the victim and inspected:
* Replicated the board in #58 in GTP mode and checked that the victim does not play in its pass-alive territory
* Ran a [long training run](https://www.notion.so/chaiberkeley/adv-b6-600-vs-avoid-pass-alive-v1-curriculum-start-at-cp39-again-ba7cb9bd348d409db3e6fb9c89a56227) where the victim is pass-hardened and manually inspected a few games to check they don't look weird
* Launch some match games and check nothing crashes: `/engines/KataGo-custom/cpp/katago match -config /go_attack/configs/match.cfg -config /go_attack/configs/compute/1gpu.cfg -override-config nnModelFile=/models/victim/cp63.txt.gz -override-config maxVisits0=1 -override-config maxVisits1=2 -override-co
nfig passingBehavior=avoid-pass-alive-territory -override-config searchAlgorithm1=MCTS -override-config logSearchInfo=false`

Also:
* Ran [some `match`es](https://www.notion.so/chaiberkeley/match-adv-600-vs-cp505-v-1-512-avoid-pass-alive-ab7e0021712146d18c552d4566685915) to check that 
  * avoid-pass-alive victims are not significantly worse than unhardened victims when they play each other
  * our adversaries that beat unhardened victims with the pass-trick fail to beat avoid-pass-alive victims, as expected
* Ran [a `match`](https://www.notion.so/chaiberkeley/match-cp63-v1-avoid-pass-alive-patch-vs-cp63-v1-avoid-pass-alive-62331e5c7ecd40e9ab042214318d2353) to check that avoid-pass-alive cp63-v1 with this PR's changes are not significantly worse than avoid-pass-alive cp63-v1 without this PR's changes when they play each other


cpp/search/searchresults.cpp
```diff
@@ -204,15 +223,23 @@ bool Search::getPlaySelectionValues(
...
+    numChildren = 0;
+    if(retVisitCounts != NULL)
+      (*retVisitCounts).clear();
 
     bool obeyAllowedRootMove = true;
     while(true) {
       for(int movePos = 0; movePos<policySize; movePos++) {
         Loc moveLoc = NNPos::posToLoc(movePos,rootBoard.x_size,rootBoard.y_size,nnXLen,nnYLen);
-        if(suppressPass && moveLoc == Board::PASS_LOC) {
+        const bool isSuppressedMove = (suppressPass && moveLoc == Board::PASS_LOC)
```
ed1d1a8d: I think there is a potential failure mode here where there are legal moves outside of {passing, pass-alive territory}, but our NN assigns near zero probability (i.e. < 1e-50) to those moves (maybe because it was trained poorly). In this case we would return no legal moves and I think KataGo will crash. Maybe we can adjust the logic so that if there are literally no legal moves, we default to passing?

It'd be nice if we could add a test for this case as well.

Let me know if this makes sense or I am just being mistakenly paranoid.

ed1d1a8d (changes requested):
Sorry for taking so long to get around to this.

Overall a really comprehensive PR, thanks for running all the tests to cross-check that this doesn't break anything. Also the fact that we got our circler-adversary working against this PR is strong evidence that it works.

Just had one requested change to handle what I think is an edge case we don't currently cover.

cpp/search/searchresults.cpp
tomtseng: In practice this hasn't been an issue — our hardened training run hasn't crashed due to this.

TBH not sure how to run KataGo tests but I've added some logic to handle the failure mode

AdamGleave:
> TBH not sure how to run KataGo tests but I've added some logic to handle the failure mode

The TL;DR is to run the scripts in `engines/KataGo-custom/cpp/run*tests*.sh` and then do a `git diff` of the output. However, @norabelrose has found some anomalies here even on our stable branch, so worth syncing up with her.

@ed1d1a8d are you OK handling the re-review? Let me know if there's anything you'd like me to take a look at.

norabelrose (approved):
This looks good to me, I just ran `runsearchtests.sh` and there's no diff in the output relative to `stable`.

cpp/search/searchresults.cpp
```diff
@@ -204,15 +223,23 @@ bool Search::getPlaySelectionValues(
   //If we have no children, then use the policy net directly. Only for the root, though, if calling this on any subtree
   //then just require that we have children, for implementation simplicity (since it requires that we have a board and a boardhistory too)
   //(and we also use isAllowedRootMove and avoidMoveUntilByLoc)
-  if(numChildren == 0) {
+  //We also use this branch if all of our children correspond to suppressed moves.
+  if(maxUnsuppressedChildWeight == 0.0) {
```
ed1d1a8d: Should we compute `maxUnsuppressedChildWeight` right before this check?

My motivations:
1. It makes the code clearer. We don't have to think about `maxUnsuppressedChildWeight` as we are reading code until right before we need it.
2. We actually modify `playSelectionValues` even after `maxUnsuppressedChildWeight` is computed. So `maxUnsuppressedChildWeight` is actually some weird stale value.

Finally, what do you think about renaming `maxUnsuppressedChildWeight` to `maxPlaySelectionValue`?

cpp/search/searchresults.cpp
```diff
@@ -87,12 +98,17 @@ bool Search::getPlaySelectionValues(
     totalChildWeight += childWeight;
     if(childWeight > maxChildWeight)
       maxChildWeight = childWeight;
-    if(suppressPass && moveLoc == Board::PASS_LOC) {
+    const bool isSuppressedMove = (suppressPass && moveLoc == Board::PASS_LOC)
```
ed1d1a8d: Could we refactor this out into a helper lambda function? We have it written in three different places and I think it would be less error prone if we didn't repeat ourselves.

ed1d1a8d (changes requested):
The fallback pass option looks good.

Just requested a few more small refactors to make things cleaner and less error prone.

I might actually take a shot at writing up a small test of this myself.

tomtseng:
Addressed comments. If you don't end up having time to write a test then I'll invest some time in figuring out how to run/write tests

ed1d1a8d:
> Addressed comments

LGTM!

> If you don't end up having time to write a test then I'll invest some time in figuring out how to run/write tests

I added some tests and they will fail if we don't have the extra failure mode logic you added. If the tests look good to you am fine with merging.

ed1d1a8d (approved):

cpp/main.cpp
```diff
@@ -48,6 +48,7 @@ evalsgf : Utility/debug tool, analyze a single position of a game from an SGF fi
 
 runtests : Test important board algorithms and datastructures
 runnnlayertests : Test a few subcomponents of the current neural net backend
+runexpensivetests : Expensive tests (mostly for the go_attack project).
```
ed1d1a8d, line 51: Added this cause we forgot to update the command line help string when we added `runexpensivetests` earlier.

PR #63, opened 2022-11-05
summarize_sgfs: Allow filtering by board size

tomtseng:
Sometimes I use `summarize_sgfs.py` to look at the win rate of the latest model in a training run, but I only want to see 19x19 boards. This PR adds a `-board-size 19` flag that makes `summarize_sgfs.py` filter to 19x19 boards.


AdamGleave (approved):
LGTM

PR #64, opened 2022-11-07
Support predictor networks in evaluate_loop.sh

norabelrose:
Mostly self-explanatory. I did remove the `SLEEP_INTERVAL` command line option on `evaluate_loop.sh` because it wasn't being used in other code and would have made the argument parsing more annoying/complicated to continue supporting it alongside the new `PREDICTOR_DIR` argument.

### DIFF
--- a/cpp/evaluate_loop.sh
+++ b/cpp/evaluate_loop.sh
@@ -1,5 +1,5 @@
 #!/bin/bash -eu
-if [[ $# -lt 1 || $# -gt 3 ]]
+if [[ $# -lt 1 || $# -gt 2 ]]
 then
     echo "Usage: $0 BASE_DIR [PREDICTOR_DIR]"
     echo "Currently expects to be run from within the 'cpp' directory of the KataGo repo."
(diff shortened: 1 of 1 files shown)


cpp/evaluate_loop.sh
```diff
@@ -48,13 +49,20 @@ do
             if [ "$STEP" -gt "$LAST_STEP" ]; then
                 # https://stackoverflow.com/questions/12152626/how-can-i-remove-the-extension-of-a-filename-in-a-shell-script
                 VICTIM_NAME=$(echo "$VICTIM" | cut -f 1 -d '.')
+                EXTRA_CONFIG="numGamesTotal=100"
+
+                if [ -n "$PREDICTOR_DIR" ]; then
+                    # https://stackoverflow.com/questions/4561895/how-to-recursively-find-the-latest-modified-file-in-a-directory
+                    PREDICTOR=$(find $PREDICTOR_DIR -name *.bin.gz -type f -printf '%T@ %p\n' | sort -n | tail -1 | cut -f2- -d" ")
```
tomtseng, line 56: nit: maybe make `$PREDICTOR_DIR` be  `"$PREDICTOR_DIR"/models` instead so that we're being more granular with the search?

I feel like there should be a way to make this simplier by `ls -v` in `"$PREDICTOR_DIR"/models` and getting the last line, but I guess it's not a big deal

tomtseng (approved):

PR #67, opened 2022-11-17
Renamed EMCTS1 -> AMCTS; factored AMCTS-R and AMCTS-S++ into separate algos

norabelrose:
We call our adversarial MCTS algorithm "A-MCTS" in the paper, so this PR updates the codebase to reflect that. In this PR, there are 4 different values for `SearchParams::SearchAlgorithm`:
1. `MCTS`
2. `AMCTS_S`, the base version of A-MCTS which evaluates the victim policy net once per node
3. `AMCTS_SXX`, or A-MCTS-S++, which averages evaluations over board symmetries
4. `AMCTS_R`, or A-MCTS-R, the full recursive version of the algorithm.

The `SearchParams` options `oppVisitsOverride` and `oppRootSymmetriesOverride` have been removed. This does represent a reduction in flexibility, since before it was theoretically possible to run search with, say, `oppVisitsOverride = 5`, thereby interpolating between A-MCTS-S and A-MCTS-R, but in practice it seems that we always set `oppVisitsOverride = 1` if we set it at all.


cpp/configs/emcts1/components/komi-handicap.cfg
```diff
@@ -1,12 +1,16 @@
 # Parameters affecting komi and handicap.
 
-# We make komi a constant 7.5, which is the most common komi for area scoring
-# per https://en.wikipedia.org/wiki/Komi_(Go)
-komiMean = 7.5
+# We make komi a constant 6.5, which is the most common komi per
+# https://en.wikipedia.org/wiki/Komi_(Go)
+# komiMean = 6.5
```
tomtseng: should we just remove this comment if we're commenting out `komiMean = 6.5`?

cpp/configs/emcts1/base.cfg
```diff
@@ -4,21 +4,18 @@
 @include components/logs.cfg
```
tomtseng, line 4: Rename the directory from `cpp/configs/emcts1` -> `cpp/configs/amcts`?

cpp/search/searchparams.h
```diff
@@ -29,17 +29,21 @@ struct SearchParams {
   bool forceWinningPass;
 
   // Algorithm to use for search
-  enum class SearchAlgorithm { MCTS, EMCTS1 };
+  enum class SearchAlgorithm {
+    // Vanilla self-play MCTS
+    MCTS,
+    // A-MCTS-S: Adversarial MCTS with sampling
```
tomtseng, line 35: Doesn't really matter but we're being a little inconsistent about whether we're writing A-MCTS (in the paper and in the comments for `enum class SearchAlgorithm` in this PR) vs. AMCTS (everywhere else in this PR)

cpp/search/search.cpp
```diff
@@ -491,22 +487,14 @@ Search::Search(
...
-      break;
-    default:
-      ASSERT_UNREACHABLE;
+  if (searchParams.usingAdversarialAlgo()) {
+    assert(searchParams.numThreads == 1); // We do not support multithreading with AMCTS (yet).
+    assert(oppNNEval != nullptr);
+    assert(!oppParams.usingAdversarialAlgo());
+
+    oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R ? oppParams.maxVisits : 1;
+    oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_SXX ? oppParams.rootNumSymmetriesToSample : 1;
```
tomtseng: this overrides  `oppParams.rootNumSymmetriesToSample` to `1` for AMCTS-R, is that what we expect? — do we want to do
```
oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
```
instead?

cpp/configs/emcts1/components/komi-handicap.cfg
norabelrose: Done

cpp/search/searchparams.h
norabelrose: Hmm yeah I think `A_MCTS_R` seems kind of clunky though? Also I kind of think it would be better to spell it "AMCTS" in the paper

cpp/search/search.cpp
norabelrose: Yeah it's actually unclear in the paper whether AMCTS-R includes the board symmetries thing or not

tomtseng (approved):

ed1d1a8d (approved):
Looks good. Feels nice to have a cleaned up codebase!

Also just made a small commit changing the last of the EMCTS's to AMCTS. Now searching the repo returns no hits on `EMCTS`.

PR #68, opened 2022-11-22
evaluate_loop: Don't error if predictor is not specified

tomtseng:
Fixes the following error that occurs if the optional `$PREDICTOR` positional argument is not provided:
* `/engines/KataGo-custom/cpp/evaluate_loop.sh: line 12: $2: unbound variable`


norabelrose (approved):
LGTM

PR #70, opened 2022-11-30
Update evaluate_loop.sh

UFO-101:
Update `evaluate_loop.sh` so that victim and adversary names are set to the same format as in training. Add more options including `VICTIMS_LIST` which enables running against multiple victims.

Ticket: https://www.notion.so/chaiberkeley/evaluate_loop-sh-change-victim-and-adversary-name-to-include-metadata-and-add-support-for-evaluati-70d91ac930e043bbac4f85e4765027c1

### DIFF
--- a/cpp/evaluate_loop.sh
+++ b/cpp/evaluate_loop.sh
@@ -1,25 +1,72 @@
 #!/bin/bash -eu
-if [[ $# -lt 1 || $# -gt 5 ]]
-then
-    echo "Usage: $0 BASE_DIR OUTPUT_DIR [KATAGO_BIN='/engines/KataGo-custom/cpp/katago GO_ATTACK_ROOT='/go_attack' VICTIM_LIST='' VICTIM_DIR='BASE_DIR/victims'  PREDICTOR_DIR='']"
+
+usage() {
+    echo "Usage: $0 [--victim-list VICTIM_LIST] [-victim-dir VICTIM_DIR]"
+    echo "          [--prediction-dir PREDICTOR_DIR] [--katago-bin KATAGO_BIN]"
+    echo "          [--go-attack-root GO_ATTACK_ROOT] BASE_DIR OUTPUT_DIR"
+    echo
+    echo "positional arguments:"
+    echo " BASE_DIR The root of the training run, containing selfplay data, models and"
+    echo " related directories."
+    echo " OUTPUT_DIR The directory to output results to."
+    echo
+    echo "optional arguments:"
+    echo "  -v VICTIM_LIST, --victim-list VICTIM_LIST"
+    echo "    A comma-separated list of models in VICTIM_DIR to use."
+    echo "    If empty, the most recent model will be used."
+    echo "  -d VICTIM_DIR, --victim-dir VICTIM_DIR"
+    echo "    The directory containing the victim models."
+    echo "    Default: BASE_DIR/victims"
+    echo "  -p PREDICTOR_DIR, --prediction-dir PREDICTOR_DIR"
+    echo "    The path containing predictor models, if applicable."
+    echo "  -k KATAGO_BIN, --katago-bin KATAGO_BIN"
+    echo "    The path to the KataGo binary."
+    echo "    Default: '/engines/KataGo-custom/cpp/katago'"
+    echo "  -g GO_ATTACK_ROOT, --go-attack-root GO_ATTACK_ROOT"
+    echo "    The root directory of the go-attack repository."
+    echo "    Default: '/go_attack'"
+    echo
+    echo "Optional arguments should be specified before positional arguments."
     echo "Currently expects to be run from within the 'cpp' directory of the KataGo repo."
-    echo "BASE_DIR is the root of the training run, containing selfplay data, models and related directories."
-    echo "VICTIM_LIST is a space or comma-separated list of models in VICTIM_DIR. If empty the most recent model will be used."
-    echo "PREDICTOR_DIR is the path containing predictor models, if applicable."
-    exit 0
+}
+
+VICTIM_LIST=""
+VICTIMS_DIR=""
+PREDICTOR_DIR=""
+KATAGO_BIN="/engines/KataGo-custom/cpp/katago"
+GO_ATTACK_ROOT="/go_attack"
+
+# Command line flag parsing (https://stackoverflow.com/a/33826763/4865149)
+while true; do
+  case $1 in
+    -h|--help) usage; exit 0 ;;
+    -v|--victim-list) VICTIM_LIST="$2"; shift 2 ;;
+    -d|--victim-dir) VICTIMS_DIR="$2"; shift 2 ;;
+    -p|--prediction-dir) PREDICTOR_DIR="$2"; shift 2 ;;
+    -k|--katago-bin) KATAGO_BIN="$2"; shift 2 ;;
+    -g|--go-attack-root) GO_ATTACK_ROOT="$2"; shift 2 ;;
+    -*) echo "Unknown parameter passed: $1"; usage; exit 1 ;;
+    *) break ;;
+  esac
+done
+
+NUM_POSITIONAL_ARGUMENTS=2
+if [ $# -ne ${NUM_POSITIONAL_ARGUMENTS} ]; then
+  echo "Wrong number of positional arguments. Expected ${NUM_POSITIONAL_ARGUMENTS}, got $#"
+  echo "Positional arguments: $@"
+  usage
+  exit 1
 fi
 
 BASE_DIR="$1"
-PREDICTOR_DIR=${2:-}
 MODELS_DIR="$BASE_DIR"/models
 OUTPUT_DIR="$2"
 mkdir -p "$OUTPUT_DIR"/logs
 mkdir -p "$OUTPUT_DIR"/sgfs
-KATAGO_BIN="${3:-'/engines/KataGo-custom/cpp/katago'}"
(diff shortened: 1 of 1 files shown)


cpp/evaluate_loop.sh
```diff
@@ -1,20 +1,25 @@
 #!/bin/bash -eu
-if [[ $# -lt 1 || $# -gt 2 ]]
+if [[ $# -lt 1 || $# -gt 5 ]]
```
tomtseng: is `-gt 5` correct? looks like we have substantially more than that many arguments

cpp/evaluate_loop.sh
```diff
@@ -1,20 +1,25 @@
...
 then
-    echo "Usage: $0 BASE_DIR [PREDICTOR_DIR]"
+    echo "Usage: $0 BASE_DIR OUTPUT_DIR [KATAGO_BIN='/engines/KataGo-custom/cpp/katago GO_ATTACK_ROOT='/go_attack' VICTIM_LIST='' VICTIM_DIR='BASE_DIR/victims'  PREDICTOR_DIR='']"
     echo "Currently expects to be run from within the 'cpp' directory of the KataGo repo."
     echo "BASE_DIR is the root of the training run, containing selfplay data, models and related directories."
+    echo "VICTIM_LIST is a space or comma-separated list of models in VICTIM_DIR. If empty the most recent model will be used."
     echo "PREDICTOR_DIR is the path containing predictor models, if applicable."
```
tomtseng: Kinda inconvenient to specify the PREDICTOR_DIR now since we'd need to specify all the other positional arguments as well. We may want to convert these from positional arguments to flags (e.g., see how `go_attack/kubernetes/launch-job.sh` handles arguments).

cpp/evaluate_loop.sh
```diff
@@ -1,20 +1,25 @@
 #!/bin/bash -eu
-if [[ $# -lt 1 || $# -gt 2 ]]
+if [[ $# -lt 1 || $# -gt 5 ]]
 then
-    echo "Usage: $0 BASE_DIR [PREDICTOR_DIR]"
+    echo "Usage: $0 BASE_DIR OUTPUT_DIR [KATAGO_BIN='/engines/KataGo-custom/cpp/katago GO_ATTACK_ROOT='/go_attack' VICTIM_LIST='' VICTIM_DIR='BASE_DIR/victims'  PREDICTOR_DIR='']"
```
tomtseng: When do we ever want to set KATAGO_BIN or GO_ATTACK_ROOT to anything else? Is it so that we can run the script outside the `goattack:cpp` container and make it easier to debug depending on the developer's setup?

cpp/evaluate_loop.sh
```diff
@@ -26,52 +31,63 @@ do
...
-    # https://stackoverflow.com/questions/1015678/get-most-recent-file-in-a-directory-on-linux
-    VICTIM=$(ls -Art "$VICTIMS_DIR" | tail -n 1)
+
+    if [[ -z "$VICTIM_LIST" ]]
+    then
+        # https://stackoverflow.com/questions/1015678/get-most-recent-file-in-a-directory-on-linux
+        VICTIM_LIST=$(ls -Art "$VICTIMS_DIR" | tail -n 1)
+    fi
+
+    # https://stackoverflow.com/a/10586169/7086623
```
tomtseng, line 88:
```suggestion
    # Split the string VICTIM_LIST into an array victim_array: https://stackoverflow.com/a/10586169/7086623
```

cpp/evaluate_loop.sh
```diff
@@ -26,52 +31,63 @@ do
...
-            # Have we evaluated this model yet?
-            if [ "$STEP" -gt "$LAST_STEP" ]; then
-                # https://stackoverflow.com/questions/12152626/how-can-i-remove-the-extension-of-a-filename-in-a-shell-script
-                VICTIM_NAME=$(echo "$VICTIM" | cut -f 1 -d '.')
-                EXTRA_CONFIG="numGamesTotal=100"
+                # Have we evaluated this model yet?
+                if [ "$STEP" -gt "$LAST_STEP" ]; then
+                    # https://stackoverflow.com/questions/12152626/how-can-i-remove-the-extension-of-a-filename-in-a-shell-script
+                    VICTIM_NAME=$(echo "$VICTIM" | cut -f 1 -d '.')
+                    EXTRA_CONFIG="numGamesTotal=100,numGameThreads=10,maxMovesPerGame=300"
```
tomtseng, line 112: why these config changes? `maxMovePerGame=300` in particular seems like it will cut a lot of games short

tomtseng (approved):

cpp/evaluate_loop.sh
AdamGleave: :+1: looks like we can have up to 7 arguments?

cpp/evaluate_loop.sh
AdamGleave: Personally fine with having extra arguments so long as the defaults are fine, being able to run things from outside container can be quite helpful for development, but some documentation on this could help (e.g. adding usage in the echo statements below)

cpp/evaluate_loop.sh
AdamGleave: :+1: flags seem better.

AdamGleave (changes requested):
Agree with Tom's comments (thanks for the review).

One additional question: will this evaluation loop be able to keep up with training? Some concern it could just fall further and further behind especially when evaluating against large enough number of victims. Might be nice to support skipping steps (just evaluate the latest model), or add some logging to warn if it is falling behind (so people can reduce # of victims or increase # of GPUs allocated to job or similar).

cpp/evaluate_loop.sh
UFO-101: Oh, oops, I was using for debugging. Removed 👍

cpp/evaluate_loop.sh
UFO-101: Using the VSCode Dev Containers setup described [here](https://github.com/HumanCompatibleAI/go_attack/pull/72), you want to have different directories because the default compilation directory is `${workspaceFolder}/build/katago` (when building with the C++ extension) and the repo root is `workspace`.

UFO-101: Added more info to flags.

cpp/evaluate_loop.sh
UFO-101: Switched to flags 👍

UFO-101:
Thanks both for your comments. I've switched to flags, following the style of `launch-job.sh`. There's also better usage info accessible through `-h/--help`.

> One additional question: will this evaluation loop be able to keep up with training?

Hmm, not really sure how to answer or address this right now, as I'm not really familiar with the training code. I actually thought this script was just for evaluation (not training). So I definitely can look into this, but it will take a bit more time.

AdamGleave (approved):
LGTM

AdamGleave:
> Thanks both for your comments. I've switched to flags, following the style of `launch-job.sh`. There's also better usage info accessible through `-h/--help`.

Thanks for making the changes, Joseph!
> Hmm, not really sure how to answer or address this right now, as I'm not really familiar with the training code. I actually thought this script was just for evaluation (not training). So I definitely can look into this, but it will take a bit more time.

OK no need to look into this now, @tomtseng can just monitor the situation in his upcoming training runs and we can parallelize it if needed.

For context, it's not used _for training_ (i.e. it has no effect on the model learned), but it is used for real-time evaluation _of the training process_. So, if evaluation is slower than training, we'll have increasingly delayed metrics.

tomtseng:
I looked at our coldstart training run just now `ls -lv /nas/ucb/k8/go-attack/victimplay/ttseng-avoid-pass-alive-coldstart-39-20221025-175949/eval/logs`, and evaluation fell behind training significantly.
* We trained adversary models faster than we evaluated then. Up to the end of October, training would output a model every <1 hour whereas evaluation finished every 4 hours. Training hit 213mil steps and evaluation hit 33mil steps by October 31.
* Then I restarted the coldstart training job to fix a bug after the curriculum hit cp505-v2. Evaluation lost its progress here because it started evaluating the adversary models from the very beginning, and it evaluated the models on the most recent and more expensive victim, cp505. Evaluation hit 21mil steps
* I think Adam restarted the coldstart training job yesterday, so now evaluation lost its progress again and is re-evaluating every adversary model on the most recent victim, cp505-v262144.

So there are a couple issues to resolve if we have extra time:
* Evaluation indeed can fall behind training even in a typical experiment.
* Restarting makes us evaluate all adversaries from the very beginning.

Maybe the answer is to have the eval loop always evaluate the latest adversary it sees rather than having it evaluate every adversary in ascending order.

AdamGleave:
> So there are a couple issues to resolve if we have extra time:
> 
> * Evaluation indeed can fall behind training even in a typical experiment.
> * Restarting makes us evaluate all adversaries from the very beginning.
> 
> Maybe the answer is to have the eval loop always evaluate the latest adversary it sees rather than having it evaluate every adversary in ascending order.

Oh, these are pretty major issues.

I do think always evaluating the latest adversary is a reasonable solution for the purpose of statistics during training. Would you be OK making a PR to do that?

We could have a more sophisticated solution that adds more parallelism to evaluation, and saves the last evaluated step to disk to allow resumes. But I think that'd only be worth it if were going to use these eval stats in final publication results rather than just for internal monitoring. But right now we have a separate workflow for that which seems like it's working OK, so probably no need to make `evaluate_loop.sh` more complex.

PR #71, opened 2022-12-14
cpp/setup: Add `oppVisitsOverride` to allow bounded A-MCTS-R

tomtseng:
* Add `oppVisitsOverride` (previously removed by PR #67) to allow bounded A-MCTS-R


AdamGleave (approved):
LGTM

PR #72, opened 2022-12-22
Merge upstream changes from lightvector/KataGo

AdamGleave:
Merge changes from upstream up to Nov 28 2022 in https://github.com/lightvector/KataGo/commit/3c42da50ec0e1dd7192ea384e9c643981b946d5e

Same as https://github.com/HumanCompatibleAI/KataGo-custom/pull/69 but with commit history tidied via squash rebase. I was only able to tidy the last part of it -- attempting to tidy earlier parts resulted in a bunch of merge conflicts. Could be done, but not automatically, as Git doesn't record the resolution of merge conflicts.

Note we will want to merge (not squash merge) this PR.

---

Notes from @tomtseng:

Notable behavior changes:
* Changes to LCB to give stronger performance at low visit counts
* Changes to FPU calculation (this can be turned off with the config param `fpuParentWeightByVisitedPolicy = false`)

I think the config changes in order to return close to v1.10.0 behavior (i.e., before this PR) are
```
useGraphSearch = false
valueWeightExponent = 0.5
subtreeValueBiasFactor = 0.35
subtreeValueBiasWeightExponent = 0.8
fpuParentWeightByVisitedPolicy = false
```

## Testing

Two tests to sanity check that the merge PRs won’t cause major issues for us:

1. Run a training run with victim visits = 1, check that we still find the passing attack. Turn gatekeeping on too to check that gatekeeper still works.
2. Run GTP games, with one player being pass-adv/cyclic-adv running on the merge PR and the other player being a victim running on `stable`. Check the win rate is similar to the same setup but with adv running on `stable` too

Results here:  [experiment page](https://www.notion.so/chaiberkeley/Nov-2022-merge-PR-v1-12-4-merge-PR-testing-af0cc4819eb043fe8451633c25fcc770)


AdamGleave:
This is a big PR that could be difficult to review. To keep things manageable, I suggest we (a) split up the work so people can review things they're specialists in, (b) use three-way diff so we can keep track of whether things changed (a) upstream, (b) in previous commits to our fork, or (c) in this PR (the thing we need to scrutinize the most).

To split up the work, I've made a [spreadsheet](https://docs.google.com/spreadsheets/d/1gZTWu3nDtMITy1xxQd2ougffT95MpyR3UF_WyHlTxeQ/edit#gid=430832183&fvid=2022518287) containing all the files and the # of LOCs changed relative to our stable branch and to KataGo upstream. Please assign yourself to files that you feel well-placed to review, and (if that's a small number) to any leftover ones. When you complete the review of the files, please tick it off on that spreadsheet.

For actually doing the reviewing, I suggest using a three way diff. One workflow that is working for me is:
  1. First add upstream, `git remote add upstream https://github.com/lightvector/KataGo.git && git fetch upstream`
  2. Make a branch to track the latest upstream commit that's been merged into this PR: `git branch katago-upstream-merge-head 3c42da50e`.
  3. `git pull` to get the latest changes from our origin, then `git checkout christmas-merge-rebased`.
  4. Do a 3-way diff between `katago-upstream-merge-head`, `HEAD` (i.e. `christmas-merge-rebased`) and `origin/stable`.

For the 3-way diff, my current preferred solution is using [diffuse](https://mightycreak.github.io/diffuse/) since it natively supports Git revisions. So you can just run `diffuse -r origin/stable -r HEAD -r katago-upstream-merge-head cpp/search/search.cpp`.

However, it's possible to use a much wider variety of tools if you grab the files manually by `git show origin/stable:cpp/search/search.cpp > cpp/search/search.cpp.stable && git show katago-upstream-merge-head:cpp/search/search.cpp > cpp/search/search.cpp.katago`. For example, in PyCharm I can just select all these files in project directory picker and hit Compare, or you could use other standalone tools like kdiff3.

cpp/core/logger.h
```diff
@@ -25,34 +29,41 @@ class Logger {
...
   bool logTime;
+  bool logConfigContents;
+  std::string header;
+
+  //------------
+
   std::vector<std::ostream*> ostreams;
   std::vector<std::ofstream*> files;
   std::vector<LogBuf*> logBufs;
   std::string logHeader;
```
AdamGleave: Do we need both this and `std::string header`?

cpp/program/setup.cpp
```diff
@@ -379,8 +378,6 @@ vector<SearchParams> Setup::loadParams(
     if(cfg.contains("searchAlgorithm"+idxStr)) params.searchAlgo = SearchParams::strToSearchAlgo(cfg.getString("searchAlgorithm"+idxStr));
     else if (cfg.contains("searchAlgorithm"))  params.searchAlgo = SearchParams::strToSearchAlgo(cfg.getString("searchAlgorithm"));
 
-    if(cfg.contains("canPassFirst"+idxStr)) params.canPassFirst = cfg.getBool("canPassFirst"+idxStr);
```
AdamGleave, line 396: Why are we removing this (config flag added as part of AMCTS)?

cpp/program/setup.cpp
```diff
@@ -367,10 +370,6 @@ vector<SearchParams> Setup::loadParams(
 
     string idxStr = Global::intToString(i);
 
-    if(cfg.contains("trackPassProb"+idxStr)) params.queryMoveLoc = cfg.getBool("trackPassProb"+idxStr) ? Board::PASS_LOC : Board::NULL_LOC;
```
AdamGleave, line 381: Is removing this intentional (as part of removing logit lens)?

python/curriculum.py
```diff
@@ -410,6 +410,13 @@ def __init__(
         logging.info("Loaded curriculum with the following params:")
         logging.info("\n".join([str(x) for x in config]))
 
+        # Make sure that all the victims exist
```
AdamGleave, line 441: This might have been better in a separate PR, but it looks fine

AdamGleave:
One high-level question I have regarding tests: are the new tests integrated into CI? Or do they rely on us manually running them and checking the diff?

I have a pretty strong preference for automating them as part of CI (even if that involves us doing a `git diff` as part of CI to check for no dirty files), otherwise I think it's inevitable we'll break it and not realize for a while.

cpp/program/setup.cpp
norabelrose: Yes

cpp/configs/misc/datamining1.cfg
```diff
@@ -33,17 +33,11 @@ cudaDeviceToUseModel0Thread0 = 0
...
 # Root move selection and biases------------------------------------------------------------------------------
 
-rootNoiseEnabled = true
-rootDirichletNoiseTotalConcentration = 10.83
-rootDirichletNoiseWeight = 0.25
+chosenMoveTemperatureEarly = 0.75
+chosenMoveTemperatureHalflife = 19
+chosenMoveTemperature = 0.15
+chosenMoveSubtract = 0
+chosenMovePrune = 1
```
tomtseng: delete these params above? (we had deleted them here: https://github.com/HumanCompatibleAI/KataGo-custom/pull/12/files#diff-8cd56b9562067171aedfe89a4ccb22a4e7c5e97db1a019755e2c2ef3eaa868c3L52 )

cpp/configs/misc/datamining4.cfg
```diff
@@ -10,3 +10,17 @@ cudaDeviceToUseModel0Thread0 = 0
...
+# Root move selection and biases------------------------------------------------------------------------------
+
+chosenMoveTemperatureEarly = 0.75
+chosenMoveTemperatureHalflife = 19
+chosenMoveTemperature = 0.15
+chosenMoveSubtract = 0
+chosenMovePrune = 1
+
+rootDesiredPerChildVisitsCoeff = 2
+rootNumSymmetriesToSample = 4
```
tomtseng: delete all changes in this file? (we deleted these params here: https://github.com/HumanCompatibleAI/KataGo-custom/pull/12/files#diff-01a3b49205a06a9cc9d4ee82937e4851032ce3e91ca1b5f17be0108389065b4cL50 )

cpp/configs/training/gatekeeper1.cfg
```diff
@@ -18,3 +18,35 @@ numNNServerThreadsPerModel = 1
 
 cudaUseFP16 = auto
 cudaUseNHWC = auto
+
+# Root move selection and biases------------------------------------------------------------------------------
```
tomtseng: delete all changes in this file? (we deleted these params here: https://github.com/HumanCompatibleAI/KataGo-custom/pull/12/files#diff-cb1f8357d255232eb1668ee8fde90780d9e4e8c62d34e54686e9968503f5754dL70 )

cpp/configs/training/gatekeeper2a.cfg
```diff
@@ -7,3 +7,38 @@ maxVisits = 200
 
 cudaDeviceToUseModel0Thread0 = 0
 cudaDeviceToUseModel0Thread1 = 1
+
+cudaUseFP16 = true
+cudaUseNHWC = true
+
+# Root move selection and biases------------------------------------------------------------------------------
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/gatekeeper2b.cfg
```diff
@@ -15,3 +15,17 @@ cudaDeviceToUseModel0Thread1 = 1
...
+
+rootEndingBonusPoints = 0.5
+rootPruneUselessMoves = true
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.1
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/gatekeeper2bfaster.cfg
```diff
@@ -21,3 +21,17 @@ cudaDeviceToUseModel0Thread0 = 0
...
+
+rootEndingBonusPoints = 0.5
+rootPruneUselessMoves = true
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.1
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/gatekeeper2c.cfg
```diff
@@ -13,3 +13,17 @@ cudaDeviceToUseModel0Thread0 = 0
...
+
+rootEndingBonusPoints = 0.5
+rootPruneUselessMoves = true
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.1
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/selfplay1.cfg
```diff
@@ -31,3 +31,20 @@ cudaUseNHWC = auto
...
+
+rootPolicyTemperatureEarly = 1.25
+rootPolicyTemperature = 1.1
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.0
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/selfplay2.cfg
```diff
@@ -30,3 +30,20 @@ cudaUseNHWC = auto
...
+
+rootPolicyTemperatureEarly = 1.25
+rootPolicyTemperature = 1.1
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.0
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg

cpp/configs/training/selfplay8a.cfg
```diff
@@ -29,3 +29,20 @@ cudaDeviceToUseModel0Thread7 = 7
...
+
+rootPolicyTemperatureEarly = 1.25
+rootPolicyTemperature = 1.1
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.0
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg

cpp/configs/training/selfplay8b.cfg
```diff
@@ -23,3 +23,48 @@ cudaDeviceToUseModel0Thread4 = 4
...
+
+rootPolicyTemperatureEarly = 1.25
+rootPolicyTemperature = 1.1
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.0
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/configs/training/selfplay8b20.cfg
```diff
@@ -27,3 +27,48 @@ cudaDeviceToUseModel0Thread4 = 4
...
+
+rootPolicyTemperatureEarly = 1.25
+rootPolicyTemperature = 1.1
+
+cpuctExploration = 1.1
+cpuctExplorationLog = 0.0
+fpuReductionMax = 0.2
+rootFpuReductionMax = 0.0
+
+numVirtualLossesPerThread = 1
```
tomtseng: same as gatekeeper1.cfg — delete all changes in this file?

cpp/dataio/trainingwrite.cpp
```diff
@@ -83,16 +89,17 @@ FinishedGameData::FinishedGameData()
    whiteValueTargetsByTurn(),
    nnRawStatsByTurn(),
 
-   queryMoveLoc(Board::NULL_LOC),
-   playoutHistoriesByTurn(),
```
tomtseng, line 87: are we getting rid of logit lens? seems like that would've been better to do in separate PR but oh well

cpp/command/selfplay.cpp
```diff
@@ -155,7 +156,8 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
 
   MakeDir::make(outputDir);
   MakeDir::make(modelsDir);
-  Logger logger(&cfg);
+
+  Logger logger;
```
tomtseng: why remove the `&cfg` argument?

cpp/core/logger.h
norabelrose: Apparently `logHeader` is never accessed ever, so I just deleted it.

cpp/configs/misc/datamining1.cfg
norabelrose: Done

cpp/configs/misc/datamining4.cfg
norabelrose: Done

cpp/configs/training/gatekeeper1.cfg
norabelrose: done

cpp/configs/training/gatekeeper2a.cfg
norabelrose: done

cpp/command/selfplay.cpp
norabelrose: I actually have no idea, just reverted

cpp/dataio/trainingwrite.cpp
norabelrose: yeahhh it was going to be hard to combine with graph search and we weren't using it

cpp/search/searchhelpers.cpp
```diff
@@ -0,0 +1,627 @@
...
+    for(int i = 0; i<numRelativeProbs; i++) {
+      //Numerically stable way to raise to power and normalize
+      processedRelProbs[i] = relativeProbs[i] <= 0.0 ? 0.0 : exp((log(relativeProbs[i]) - logMaxValue) / temperature);
+      sum += processedRelProbs[i];
+    }
+    assert(sum > 0.0);
+    uint32_t idxChosen = rand.nextUInt(processedRelProbs,numRelativeProbs);
+    return idxChosen;
+  }
+}
```
tomtseng, line 41: We modified the implementation of `chooseIndexWithTemperature` in KataGo-custom (e.g., we use `temperatureScaleProbs()` instead `chooseIndexWithTemperature` — should those modifications be reflected here?

cpp/search/search.cpp
```diff
@@ -2066,1757 +1169,211 @@ void Search::computeRootValues() {
...
-  int numLegalMovesFound = 0;
-  for(int pos = 0; pos<policySize; pos++) {
-    if(policyProbs[pos] >= 0.0f) {
-      sortedPolicyBuf[numLegalMovesFound++] = PolicySortEntry(policyProbs[pos],pos);
-    }
-  }
-  int numMovesToReturn = std::min(n,numLegalMovesFound);
-  std::partial_sort(sortedPolicyBuf,sortedPolicyBuf+numMovesToReturn,sortedPolicyBuf+numLegalMovesFound);
-  return numMovesToReturn;
+bool Search::runSinglePlayout(SearchThread& thread, double upperBoundVisitsLeft) {
```
tomtseng: we had changed this to return an `int` (`numNodesAdded`) rather than a bool, but  that's been reverted in this PR. do we not care about that change?

cpp/search/searchexplorehelpers.cpp
```diff
@@ -0,0 +1,478 @@
...
+      parentWeightPerVisit,parentUtilityStdevFactor,
+      maxChildWeight,&thread
+    );
+    if(selectionValue > maxSelectionValue) {
+      maxSelectionValue = selectionValue;
+      bestChildIdx = numChildrenFound;
+      bestChildMoveLoc = bestNewMoveLoc;
+    }
+  }
+  // cout << "GT bestChildMoveLoc: " << Location::toString(bestChildMoveLoc, thread.board) << ", PSV: " << maxSelectionValue << endl;
```
tomtseng: remove dead debugging code

cpp/search/searchresults.cpp
```diff
@@ -26,7 +31,7 @@ bool Search::shouldSuppressMove(
 ) const {
```
tomtseng: maybe: move this function to `searchhelpers.cpp` so that it's defined near the related function `shouldSuppressPass()` (which used to be in `searchresults.cpp` and got moved to `searchhelpers.cpp` in this merge)

cpp/search/searchhelpers.cpp
```diff
@@ -0,0 +1,627 @@
...
+    double leadAvg = node.stats.leadAvg.load(std::memory_order_acquire);
+    double utilityAvg = node.stats.utilityAvg.load(std::memory_order_acquire);
+
+    if(numVisits <= 0 || weightSum <= 1e-10)
+      return false;
+    passWeight = weightSum;
+    passUtility = utilityAvg;
+    passScoreMean = scoreMeanAvg;
+    passLead = leadAvg;
+  }
```
tomtseng: some changes in this block didn't get copied over from upstream? e.g., see https://github.com/lightvector/KataGo/commit/a306a77760712f67c57a3f5bd8825cf9a66b807e

cpp/search/searchresults.cpp
```diff
@@ -1893,23 +1786,18 @@ bool Search::getAnalysisJson(
...
+      moveInfo["utility"] = Global::roundDynamic(utility,OUTPUT_PRECISION);
+      moveInfo["winrate"] = Global::roundDynamic(winrate,OUTPUT_PRECISION);
+      moveInfo["scoreMean"] = Global::roundDynamic(lead,OUTPUT_PRECISION);
+      moveInfo["scoreSelfplay"] = Global::roundDynamic(scoreMean,OUTPUT_PRECISION);
+      moveInfo["scoreLead"] = Global::roundDynamic(lead,OUTPUT_PRECISION);
+      moveInfo["scoreStdev"] = Global::roundDynamic(data.scoreStdev,OUTPUT_PRECISION);
+      moveInfo["prior"] = Global::roundDynamic(data.policyPrior,OUTPUT_PRECISION);
+      moveInfo["lcb"] = Global::roundDynamic(lcb,OUTPUT_PRECISION);
+      moveInfo["utilityLcb"] = Global::roundDynamic(utilityLcb,OUTPUT_PRECISION);
+      moveInfo["order"] = data.order;
```
tomtseng, line 1792: __missing compared to upstream:__
```
moveInfo["weight"] = data.weightSum;
```
and this comment above `moveInfo["scoreMean"]`:
```
    // We report lead for scoreMean here so that a bunch of legacy tools that use KataGo use lead instead, which
    // is usually a better field for user applications. We report scoreMean instead as scoreSelfplay
```

__missing compared to our KataGo-custom changes:__
```
     moveInfo["selectionProb"] = roundDynamic(data.selectionProb,JSON_PRECISION);
     moveInfo["selectionValue"] = roundDynamic(data.playSelectionValue,JSON_PRECISION);
```
(but I'm guessing these got removed as part of the logit lens removal?)
and this comment above `moveInfo["move"]`:
```
      // This only depends on the size of the board, which is invariant, so it's fine to use the root board here
```

cpp/search/searchresults.cpp
```diff
@@ -1926,21 +1814,25 @@ bool Search::getAnalysisJson(
...
         moveInfo["pvVisits"] = pvVisits;
+
+        assert(data.pvEdgeVisits.size() >= pvLen);
+        json pvEdgeVisits = json::array();
+        for(int j = 0; j < pvLen; j++)
+          pvEdgeVisits.push_back(data.pvEdgeVisits[j]);
+        moveInfo["pvEdgeVisits"] = pvEdgeVisits;
       }
 
-      if(node == rootNode) {
```
tomtseng, line 1934: we had added this `if(node == rootNode)` guard, do we not need it anymore?

cpp/search/searchupdatehelpers.cpp
```diff
@@ -0,0 +1,476 @@
...
+}
+
+double Search::computeWeightFromNode(const SearchNode& node) const {
+  // If doing AMCTS, we weight opponent nodes as zero.
+  if (node.nextPla != rootNode->nextPla && searchParams.oppWeightZeroingOverride.value_or(searchParams.usingAdversarialAlgo())) {
+    return 0.0;
+  }
+  const NNOutput* nnOutput = node.getNNOutput();
+  return computeWeightFromNNOutput(nnOutput);
+}
```
tomtseng, line 105: (note to other people looking through this diff: this function is replacing `computeNodeWeight()` (which was defined in `search.cpp))

cpp/search/searchparams.h
```diff
@@ -49,8 +48,9 @@ struct SearchParams {
   // Overrides the number of visits we use in AMCTS-R to simulate the victim.
   std::optional<int> oppVisitsOverride;
 
-  // Whether we are allowed to be the first player that passes in a game.
-  bool canPassFirst;
+  // If non-none, determines whether to set the weight of opponent nodes to zero.
+  // By default, this will be false for MCTS and true for adversarial algorithms.
+  std::optional<bool> oppWeightZeroingOverride;
```
tomtseng, line 53: this is new to this PR. why do we need this? it's never assigned a value anywhere

cpp/tests/testamcts.cpp
```diff
@@ -350,12 +330,58 @@ void AMCTSTests::testAMCTS(const int maxVisits,
       bot.makeMove(loc, curPla);
       curPla = getOpp(curPla);
 
-      // Make sure game hasn't been prematurely ended.
-      testAssert(!bot.rootHistory.isGameFinished);
```
tomtseng: why did this assert change?

cpp/tests/testamcts.cpp
```diff
@@ -332,6 +310,8 @@ void AMCTSTests::testAMCTS(const int maxVisits,
 
     Player curPla = P_BLACK;
     for (int midx = 0; midx < numMovesToSimulate; midx++) {
+      Board::printBoard(cout, bot.rootBoard, Board::NULL_LOC, &bot.rootHistory.moveHistory);
```
tomtseng: what's a typical value of `numMovesToSimulate`? Will this printing be noisy?

cpp/tests/testamcts.cpp
```diff
@@ -598,273 +560,104 @@ void AMCTSTests::checkFinalMoveSelection(const Search& bot) {
 
   testAssert(childToPsv.size() == trueLocToPsv.size());
   for (const auto [child, psv] : childToPsv) {
-    testAssert(approxEqual(trueLocToPsv[child->prevMoveLoc], psv));
+    const Loc loc = tree.prevMoves.at(child);
+    testAssert(approxEqual(trueLocToPsv[loc], psv));
   }
 }
 
-void AMCTSTests::checkPlayoutLogic(const Search& bot) {
```
tomtseng: why did this get removed?

cpp/tests/testamcts.cpp
```diff
@@ -550,15 +509,17 @@ void AMCTSTests::checkFinalMoveSelection(const Search& bot) {
...
+      for (const auto edge : tree.outEdges.at(tree.root)) {
+        auto child = edge->getIfAllocated();
+        const Loc loc = tree.prevMoves.at(child);
+        const int64_t visits = edge->getEdgeVisits();
+        bot.getSelfUtilityLCBAndRadius(*tree.root, child, visits, loc, lcbs[child],
                                        radii[child]);
         double weight = childToPsv[child];
-        if (weight >= MIN_WEIGHT_FOR_LCB &&
-            weight >= bot.searchParams.minVisitPropForLCB * maxChildWeight) {
+        if (weight > 0 && weight >= bot.searchParams.minVisitPropForLCB * maxChildWeight) {
```
tomtseng: why'd this change from `weight >= MIN_WEIGHT_FOR_LCB` to `weight > 0`?

cpp/tests/testamcts.h
```diff
@@ -16,13 +19,16 @@ void testConstPolicies();
 // Test our modifications didn't break the original MCTS.
 void testMCTS(const int maxVisits, const int numMovesToSimulate);
 
+// Test AMCTS
+void testAMCTS(const int maxVisits, const int numMovesToSimulate);
+
+std::shared_ptr<NNEvaluator> getNNEval(std::string modelFile, ConfigParser& cfg,
+                                       Logger& logger, uint64_t seed);
```
tomtseng: why move this out of the `/** Helper functions **/` section? 
In general a lot of stuff got rearranged in `testamcts.h` and `testamcts.cpp` (e.g., `/** Constants **/` got moved from the bottom of the file to above `/** Helper functions **/`) but it seems out of scope for this merge. Not worth the effort to revert now that the rearrangement is done (assuming we prefer the rearrangement) but made it a bit more confusing to review these files

cpp/tests/testsearch.cpp
```diff
@@ -384,4114 +174,21 @@ static void runBasicPositions(NNEvaluator* nnEval, Logger& logger)
...
-    cout << json << endl;
-
-    // Test with includeTree = true
-    includeTree = true;
-    json = nlohmann::json();
-    suc = search->getAnalysisJson(
-      perspective, analysisPVLen, ownershipMinVisits, preventEncore,
-      includePolicy, includeOwnership, includeOwnershipStdev, includeMovesOwnership, includeMovesOwnershipStdev, includePVVisits, includeTree,
-      json
-    );
```
tomtseng, line 3472: We added this part where we test both `includeTrue = true` and `includeTrue = false`. This test (`"Testing pruning of search tree at root node due to symmetries, allowing only diagonal flips: empty board, with avoid moves"`) got moved to `testsearchnonn.cpp`, but we lost the part where we test both `includeTrue = {true, false}`. Should we add that back into `testsearchnonn.cpp`?

cpp/tests/testsearchnonn.cpp
```diff
@@ -0,0 +1,2277 @@
...
+    bool preventEncore = true;
+    bool includePolicy = false;
+    bool includeOwnership = false;
+    bool includeOwnershipStdev = false;
+    bool includeMovesOwnership = false;
+    bool includeMovesOwnershipStdev = false;
+    bool includePVVisits = false;
+    bool suc = search->getAnalysisJson(
+      perspective, analysisPVLen, preventEncore,
+      includePolicy, includeOwnership, includeOwnershipStdev, includeMovesOwnership, includeMovesOwnershipStdev, includePVVisits, false,
```
tomtseng: nit: throughout this file, it would be more readable to define `bool includeTree = false;` (just like how all the other arguments passed into `getAnalysisJson` are given names) instead of directly passing in `false` as an argument to `getAnalysisJson()`

tomtseng (commented):
(still need to review `cpp/tests/results` changes, after that we've reviewed everything in the spreadsheet)
EDIT: I also should re-review gatekeeper.cpp — there have been new merge conflicts there

cpp/command/gatekeeper.cpp
```diff
@@ -275,6 +276,24 @@ namespace {
...
+  // Or was it contained in a subdirectory
+  else if(Global::isPrefix(FileUtils::weaklyCanonical(modelDir), FileUtils::weaklyCanonical(testModelsDir))) {
+    string renameDest = intoDir + "/" + modelName;
+    logger.write("Moving " + modelDir + " to " + renameDest);
+    FileUtils::rename(modelDir,renameDest);
+  }
+  else {
+    throw StringError("Model " + modelDir + " does not appear to be a subdir of " + testModelsDir + " can't figure out where how to move it to accept or reject it");
+  }
+}
```
tomtseng, line 291: Compiler gives a warning that this function is unused. We should either delete it, or use it like upstream uses it — it looks like upstream replaces lines like
```cpp
      logger.write("Moving " + testModelDir + " to " + renameDest);
      FileUtils::rename(testModelDir,renameDest);
```
with
```cpp
      moveModel(testModelName, testModelFile, testModelDir, testModelsDir, rejectedModelsDir, logger);
```

So if we want to use it, search for instances of `FileUtils::rename(` in this file and replace them

cpp/command/gatekeeper.cpp
```diff
@@ -49,6 +49,7 @@ namespace {
     const SearchParams searchParamsCandidate;
     MatchPairer* matchPairer;
 
+    string testModelFile;
```
tomtseng: we're not using this (and we're not even populating it in the constructor right now — maybe a merge conflict mistake?), should remove it unless we need it for filling out arguments for `moveModel()`

tomtseng (commented):
Done reviewing all files in the spreadsheet. I also fixed a merge conflict in `setup.cpp` that was making the CI GPU tests fail

cpp/search/searchexplorehelpers.cpp
norabelrose: yep

cpp/search/searchparams.h
norabelrose: IIRC the original reason was for testing purposes, so that we could run compare vanilla MCTS to AMCTS where the victim network is identical to the adversary network. But I think that test hasn't been written (yet)

cpp/tests/testamcts.cpp
norabelrose: IIRC it was because we couldn't get it to work and it wasn't clear that the test was actually necessary; we were tentatively thinking of shifting over to a testing scheme where you just compare the logs generated by the current commit to logs generated on a past commit

cpp/tests/testamcts.cpp
norabelrose: Pretty sure upstream `MIN_WEIGHT_FOR_LCB` has been replaced by zero everywhere, in part to make the algorithm more robust to our passing attack.

cpp/command/gatekeeper.cpp
norabelrose: Yep lets use it like uptream

cpp/tests/testsearchnonn.cpp
norabelrose: sounds good

cpp/command/gatekeeper.cpp
norabelrose: yep probably a merge mistake

cpp/tests/testamcts.cpp
norabelrose: IIRC yes it is kinda noisy

cpp/tests/testamcts.cpp
norabelrose: It's really hard to remember exactly but there may have been something going on here where the old code assumed invariants that did hold with tree search but fails to hold in graph search

cpp/search/searchresults.cpp
norabelrose: ok

cpp/tests/testamcts.cpp
ed1d1a8d: Added back in.

cpp/search/search.cpp
ed1d1a8d: Has been changed to int again.

cpp/tests/testamcts.cpp
ed1d1a8d: Assert moved back in.

ed1d1a8d:
Fixed the AMCTS tests so they are working again (and modified `runSinglePlayout` to bring back the if block that was deleted).

cpp/search/search.cpp
```diff
@@ -1169,13 +1174,49 @@ void Search::computeRootValues() {
...
+
+    // Update stats along path to root (from the first playout)
+    for (int i = playoutPath.size() - 2; i >= 0; i--) {
+      SearchNode* node = playoutPath[i];
+
+      int childrenCapacity;
+      SearchChildPointer* children = node->getChildren(childrenCapacity);
+      for (int j = 0; j < childrenCapacity; j++) {
+        if (children[j].getIfAllocated() == playoutPath[i + 1]) {
+          children[j].addEdgeVisits(1);
```
tomtseng, line 1220: this is new, why was it added? I'm not familiar with this code

cpp/tests/testamcts.cpp
```diff
@@ -79,26 +78,56 @@ static void setSimpleSearchParams(SearchParams& params) {
   testAssert(!params.useNoisePruning);
   testAssert(!params.useUncertainty);
   testAssert(!params.antiMirror);
+  testAssert(!params.forceWinningPass);  // We aren't testing pass sharpening.
 }
 
+// Almost directly copied from Search::getFpuValueForChildrenAssumeVisited.
+// Irrelevant code removed.
```
tomtseng: Maybe also explain in this comment why we need to have this function instead of just using `Search::getFpuValueForChildrenAssumeVisited` — we want to pass in the new argument `stats` because...

tomtseng:
Forgot to ping here earlier but this should be ready for re-review @ed1d1a8d @AdamGleave 

Still doing some testing on the side, I'll update the PR description & comment here with relevant testing results later

cpp/tests/results/analysis/badoverride.txt.log
```diff
@@ -96,7 +96,7 @@ D17 : T  -1.10c W  -1.55c S   0.15c ( -0.1 L  -0.0) LCB  -21.40c P  8.00% WF   6
...
...nValue":3.0,"utility":-0.0118601734,"utilityLcb":-0.296880549,"visits":5,"weight":5.0,"winrate":0.48882949},{"lcb":0.338129367,"move":"Q16","order":11,"prior":0.0390467197,"pv":["Q16","D3","D17"],"resultUtility":0.0153946315,"scoreLead":-0.119253183,"scoreMean":-0.119253183,"scoreSelfplay":-0.204626532,"scoreStdev":30.6883878,"selectionProb":0.00954149996,"selectionValue":3.0,"utility":-0.00614373591,"utilityLcb":-0.437829025,"visits":4,"weight":4.0,"winrate":0.492302684},{"lcb":0.229640739,"move":"C3","order":12,"prior":0.0324412882,"pv":["C3","R16","C16"],"resultUtility":0.0113410552,"scoreLead":-0.664291049,"scoreMean":-0.664291049,"scoreSelfplay":-0.483069718,"scoreStdev":30.7822067,"selectionProb":0.00424066665,"selectionValue":2.0,"utility":-0.0139643066,"utilityLcb":-0.755092761,"visits":3,"weight":3.0,"winrate":0.494329472}],"rootInfo":{"currentPlayer":"B","rawStScoreError":-1.0,"rawStWrError":-0.5,"rawVarTimeLeft":-1.0,"scoreLead":0.1121465,"scoreSelfplay":0.0699747314,"scoreStdev":30.5698449,"symHash":"47E1EBDAE69A5DED3FF8DB1FA4E91845","thisHash":"47E1EBDAE69A5DED3FF8DB1FA4E91845","utility":-4.83761537e-05,"visits":100,"weight":100.0,"winrate":0.497650216},"turnNumber":0}
```
ed1d1a8d, line 99: What is the expected change for this file?

cpp/dataio/loadmodel.cpp
```diff
@@ -21,7 +21,8 @@ static const vector<string> ACCEPTABLE_MODEL_SUFFIXES {
   ".bin.gz",
   ".bin",
   "model.txt.gz",
-  "model.txt"
+  "model.txt",
+  ".txt.gz",
```
ed1d1a8d, line 23: Why do we need to add .txt.gz here?

cpp/dataio/loadmodel.cpp
```diff
@@ -64,9 +65,6 @@ bool LoadModel::findLatestModel(const string& modelsDir, Logger& logger, string&
   gfs::path latestPath;
   for(const auto& dirEntry: gfs::recursive_directory_iterator(gfs::u8path(modelsDir))) {
     gfs::path filePath = dirEntry.path();
-    if(checkDirsOnly && !gfs::is_directory(filePath))
```
ed1d1a8d: Why do we not need `checkDirsOnly` and this check anymore?

cpp/search/search.cpp
```diff
@@ -114,19 +114,23 @@ Search::Search(
...
-    oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
-      ? params.oppVisitsOverride.value_or(oppParams.maxVisits)
-      : 1;
-    oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
-    oppBot = make_unique<Search>(oppParams, oppNNEval, logger, rSeed + "-victim-model");
+    // We only want to initialize oppBot if maxVisits > 1 or else we'll get
+    // infinite recursion when both bots are using A-MCTS.
+    if (searchParams.maxVisits > 1) {
+      assert(oppNNEval != nullptr);
+      assert(!oppParams.useGraphSearch);
```
ed1d1a8d: Why can't the opponent use graph search?

cpp/search/searchresults.cpp
```diff
@@ -1762,6 +1748,9 @@ bool Search::getAnalysisJson(
     static constexpr int minMoves = 0;
 
     json moveInfos = json::array();
+    if(node == NULL) {
+      return moveInfos;
+    }
```
ed1d1a8d: Why did this need to be added?

ed1d1a8d (approved):
LGTM just had a few questions.

cpp/tests/results/analysis/badoverride.txt.log
tomtseng: In `Search::getAnalysisJson()` we have the following changes relative to upstream:
* Add `selectionProb`, `selectionValue`, and `resultUtility` fields
* Add `children` field if `includeTree` flag is used (not relevant for this file)

So we expect the addition of those fields. Maybe in order to make the diffs nicer we should just remove those fields?  `selectionProb` was introduced by the logit lens PR (#66) so we can remove it since we're not using the logit lens, but I'm not sure about `selectionValue` and `resultUtility` (introduced in PR #65, but that PR's description doesn't provide any context about why we were interested in these values)

cpp/dataio/loadmodel.cpp
tomtseng: `findLatestModel()` won't find models like `kata1-b6c96-s175395328-d26788732.txt.gz` if we don't add `.txt.gz` here. Not sure why it isn't already included in this array.

cpp/dataio/loadmodel.cpp
tomtseng: The outer `for` loop used to be a `directory_iterator`, which would non-recursively look at files inside `modelsDir` for models (if `checkDirsOnly == false`) or directories non-recursively containing models.

Now the outer `for` loop is a `recursive_directory_iterator` which recursively looks for files inside `modelsDir`, and it directly looks for models matching the list of `ACCEPTABLE_MODEL_SUFFIXES`. In particular, now if we reject any file that isn't a directory, we will never find a model since a model is always a regular file and not a directory.

cpp/search/search.cpp
tomtseng: I was following the assertion from this commit:
https://github.com/AlignmentResearch/KataGo-custom/pull/72/commits/4b603d0d89caf8039f3c6997f3fef90f04f1a90e#diff-d20a00509040ee72ce58635f85b6ca9fcabf064b3a7178f808aecbfe5e3a9b45R120-R123
```cpp
    // When using AMCTS, we don't support graph search for ourselves or our
    // opponent.
    assert(!searchParams.useGraphSearch);
    assert(!oppParams.useGraphSearch);
```
I'm not familiar enough with how A-MCTS interacts with graph search, I can remove the assert if it seems unnecessary to you

cpp/search/searchresults.cpp
tomtseng: hmm this is actually an issue that also exists on `stable` so I'll move it out of this merge PR into a separate PR.
The issue was that `./runoutputtests.sh` would crash early due to our refactor of `getAnalysisJson()` not handling `nullptr`s

cpp/tests/results/analysis/badoverride.txt.log
ed1d1a8d: Do we expect these logs to differ from KataGo upstream if we remove selectionProb, selectionValue, and resultUtility? If not, could we gate those values behind a flag that is only active on the relevant tests, so that we can double check all the logs match?

cpp/search/search.cpp
ed1d1a8d: This is to mirror the update logic done in `playoutDescend`, which updated in the new KataGo code:
https://github.com/AlignmentResearch/KataGo-custom/blob/3c42da50ec0e1dd7192ea384e9c643981b946d5e/cpp/search/search.cpp#L1316-L1325

Compare to the old code:
https://github.com/AlignmentResearch/KataGo-custom/blob/22a5ac62a0d256417fabeec81b130ad09b1c05d1/cpp/search/search.cpp#L3863-L3867

---
If we do not do this, we cannot get the AMCTS tests to work properly.

cpp/tests/testamcts.cpp
ed1d1a8d: Have improved the `getFpuValueForChildrenAssumeVisited` implementation and documentation.

cpp/search/search.cpp
ed1d1a8d: Ah I forgot this was me haha 😛 
Turns out this was me just overly conservative when getting AMCTS tests to pass.
I've now disabled this check and made AMCTS tests also test the opponent with graph search.

cpp/tests/results/analysis/badoverride.txt.log
ed1d1a8d: Ok never mind even hiding these values doesn't make the diffs of these output files nicer. https://github.com/AlignmentResearch/KataGo-custom/compare/tony/christmas-merge-rebased-json-flag..3c42da50ec0e1dd7192ea384e9c643981b946d5e. I think we can just leave as is.

ed1d1a8d:
@tomtseng Made some cleanup. The PR looks good to me, though am curious about the results of the additional tests you said you were running.

tomtseng:
Two tests to sanity check that the merge PRs won’t cause major issues for us:

1. Run a training run with victim visits = 1, check that we still find the passing attack. Turn gatekeeping on too to check that gatekeeper still works.
2. Run GTP games, with one player being pass-adv/cyclic-adv running on the merge PR and the other player being a victim running on `stable`. Check the win rate is similar to the same setup but with adv running on `stable` too

Results here:  [experiment page](https://www.notion.so/chaiberkeley/Nov-2022-merge-PR-v1-12-4-merge-PR-testing-af0cc4819eb043fe8451633c25fcc770)

PR #73, opened 2022-12-24
evaluate_loop: Only evaluate the latest model

tomtseng:
Issues with evaluate_loop (with an example given in https://github.com/HumanCompatibleAI/KataGo-custom/pull/70#issuecomment-1361865618)
* Evaluation indeed can fall behind training even in a typical experiment.
* Restarting makes us re-evaluate all adversaries

Changes:
* Make evaluate_loop only evaluate the latest adversary, not every adversary available

### DIFF
--- a/cpp/evaluate_loop.sh
+++ b/cpp/evaluate_loop.sh
@@ -37,7 +37,7 @@ KATAGO_BIN="/engines/KataGo-custom/cpp/katago"
 GO_ATTACK_ROOT="/go_attack"
 
 # Command line flag parsing (https://stackoverflow.com/a/33826763/4865149)
-while true; do
+while [ -n "${1-}" ]; do
   case $1 in
     -h|--help) usage; exit 0 ;;
     -v|--victim-list) VICTIM_LIST="$2"; shift 2 ;;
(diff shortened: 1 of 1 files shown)


cpp/evaluate_loop.sh
```diff
@@ -37,7 +37,7 @@ KATAGO_BIN="/engines/KataGo-custom/cpp/katago"
 GO_ATTACK_ROOT="/go_attack"
 
 # Command line flag parsing (https://stackoverflow.com/a/33826763/4865149)
-while true; do
+while [ -n "${1-}" ]; do
```
tomtseng, line 40: the `-u` in `#!/bin/bash -eu` means that if we don't check for `$1` here, then we get an error about `$1` being unbound in the line below

cpp/evaluate_loop.sh
```diff
@@ -85,56 +85,52 @@ do
...
-                        -override-config nnModelFile0="$VICTIMS_DIR"/"$VICTIM" \
-                        -override-config botName0="victim-$VICTIM_NAME" \
-                        -override-config nnModelFile1="$MODELS_DIR"/"$MODEL_DIR"/model.bin.gz \
-                        -override-config botName1="adv-$MODEL_DIR" \
-                        -sgf-output-dir "$OUTPUT_DIR"/sgfs/"$VICTIM_NAME"_"$MODEL_DIR" \
-                        2>&1 | tee "$OUTPUT_DIR"/logs/"$VICTIM_NAME"_"$MODEL_DIR".log
-
-                    # Update the last step
-                    LAST_STEP="$STEP"
+        if [[ "$LATEST_MODEL_DIR" =~ -s([0-9]+) ]]; then
```
tomtseng, line 101: this chunk below is just replaces `MODEL_DIR` with `LATEST_MODEL_DIR` and shifts the indentation level down by 1

UFO-101 (approved):
LGTM.
Btw could we replace these bash scripts with python (using `subprocess.run` where necessary)? I find bash very difficult to work with.

tomtseng:
yeah I find bash to be quite unwieldly. I wouldn't be opposed to writing Python scripts instead, but with enough StackOverflow I find bash to be OK enough that I don't feel compelled to actively put effort into replacing existing bash scripts with Python scripts

PR #74, opened 2022-12-26
gatekeeper: Make gatekeeping work for victimplay

tomtseng:
## Description

Context: The gatekeeper works for selfplay by taking candidate models, playing them against the latest "accepted" model (where the first accepted model is a random `/dev/null` model), and accepting a candidate model if it wins more than the accepted model.

For victimplay, we don't want the gatekeeper to play two adversaries against each other. Instead, we want to play adversaries against the latest victim and compare how good they are at beating the victim.

Changes:
* Add `victimplaygatekeeper` KataGo command that tests candidate models by playing it against the latest victim and checking if it performs better than the accepted model does against the latest victim. 

## Testing

* sanity check with some manual testing that the selfplay gatekeeper still works:
```sh
docker run --gpus '"device=7"' -v ~/go_attack:/go_attack  -v ~/gatekeep-scratch:/out  -it humancompatibleai/goattack:cpp
./engines/KataGo-custom/cpp/katago gatekeeper \
  -config /go_attack/configs/amcts/gatekeeper.cfg  \
  -config /go_attack/configs/compute/1gpu.cfg \
  -override-config numBots=1,maxVisits0=1 \
  -rejected-models-dir /out/rejected/ \
  -accepted-models-dir /out/accepted \
  -test-models-dir /out/test/ \
  -sgf-output-dir /out/output/ \
  -selfplay-dir /out/selfplay/ 
# Then move some models into ~/gatekeep-scratch/test and make sure models
# are played against each other, accepted, and rejected
```
* Do some similar manual testing to check that the victimplay gatekeeper executes as expected when new test models appear, new victim models appear, and when the victim config file changes
* Run victimplay training run with gating (go_attack PR: https://github.com/HumanCompatibleAI/go_attack/pull/84), check gatekeeper works

### DIFF
--- a/cpp/command/gatekeeper.cpp
+++ b/cpp/command/gatekeeper.cpp
@@ -250,7 +250,7 @@ namespace {
   };
 
   // Returns info about latest model in a directory.
-  // (Wrapper for  LoadModel::findLatestModel().)
+  // (Wrapper for LoadModel::findLatestModel().)
   optional<ModelFileInfo> getLatestModelInfo(
       Logger& logger,
       const string& modelsDir,
@@ -532,9 +532,9 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
   auto gameLoopProtected = [&logger,&gameLoop](int threadIdx) {
     Logger::logThreadUncaught("game loop", &logger, [&](){ gameLoop(threadIdx); });
   };
-  // Runs NetAndStuff. Returns true if the program should continue executing.
+  // Runs netAndStuff games. May quit early, in which case `shouldStop` will be
+  // true.
   const auto evaluateNetAndStuff = [
-    quitIfNoNetsToTest,
     &netAndStuff,
     &netAndStuffDataIsWritten,
     &dataWriteLoopProtected,
@@ -543,17 +543,11 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
     &logger,
     numGameThreads,
     &gameLoopProtected
-  ]() -> bool {
-    if(netAndStuff == NULL) {
-      if(quitIfNoNetsToTest) {
-        shouldStop.store(true);
-        return false;
-      }
-      return true;
-    }
+  ]() {
+    assert(netAndStuff != NULL);
     //Check again if we should be stopping, after loading the new net, and quit more quickly.
     if(shouldStop.load()) {
-      return false;
+      return;
     }
     netAndStuffDataIsWritten = false;
     logger.write(
@@ -587,7 +581,6 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
         waitNetAndStuffDataIsWritten.wait(lock);
       }
     }
-    return !shouldStop.load();
   };
 
   // Victimplay-only variables
@@ -612,6 +605,8 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
       if (rejectOldTestModel(*testModelInfo, *acceptedModelInfo)) {
         testModelInfo.reset();
       }
+    } else if (quitIfNoNetsToTest) {
+      break;
     }
 
     bool shouldAcceptTestModel = false;
@@ -655,11 +650,9 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
       if (!acceptedModelMatchup.isSameMatchup(lastAcceptedModelResults)) {
         // Need to re-evaluate accepted model vs. victim model.
         netAndStuff = loadNetAndStuff(*victimModelInfo, *acceptedModelInfo, !victimplay);
-        assert(netAndStuff != NULL);
-
         logger.write("Evaluating accepted model");
-        const bool evaluationShouldContinue = evaluateNetAndStuff();
-        if (!evaluationShouldContinue) {
+        evaluateNetAndStuff();
+        if (shouldStop.load()) {
           break;
         }
 
@@ -678,9 +671,6 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
         );
         delete netAndStuff;
         netAndStuff = NULL;
-        if (!evaluationShouldContinue) {
-          break;
-        }
       }
 
       if (!testModelInfo.has_value()) {
@@ -688,10 +678,9 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
         continue;
       }
       netAndStuff = loadNetAndStuff(*victimModelInfo, *testModelInfo, !victimplay);
-      assert(netAndStuff != NULL);
       logger.write("Evaluating test model");
-      const bool evaluationShouldContinue = evaluateNetAndStuff();
-      if (!evaluationShouldContinue) {
+      evaluateNetAndStuff();
+      if (shouldStop.load()) {
         break;
       }
 
@@ -732,8 +721,8 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
         continue;
       }
       netAndStuff = loadNetAndStuff(*acceptedModelInfo, *testModelInfo, !victimplay);
-      const bool evaluationShouldContinue = evaluateNetAndStuff();
-      if (!evaluationShouldContinue) {
+      evaluateNetAndStuff();
+      if (shouldStop.load()) {
         break;
       }
       logger.write(
--- a/cpp/command/gatekeeper.cpp
+++ b/cpp/command/gatekeeper.cpp
@@ -296,7 +296,7 @@ int MainCmds::gatekeeper(const vector<string>& args, bool victimplay) {
   try {
     KataGoCommandLine cmd("Test neural nets to see if they should be accepted for self-play training data generation.");
     cmd.addConfigFileArg("","");
-    cmd.addOverrideConfigArg()
+    cmd.addOverrideConfigArg();
 
     TCLAP::ValueArg<string> testModelsDirArg("","test-models-dir","Dir to poll and load models from",true,string(),"DIR");
     TCLAP::ValueArg<string> sgfOutputDirArg("","sgf-output-dir","Dir to output sgf files",true,string(),"DIR");
--- a/cpp/command/gatekeeper.cpp
+++ b/cpp/command/gatekeeper.cpp
@@ -254,10 +254,11 @@ namespace {
   optional<ModelFileInfo> getLatestModelInfo(
       Logger& logger,
       const string& modelsDir,
-      bool allowRandomNet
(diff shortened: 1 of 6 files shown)


cpp/program/setup.cpp
```diff
@@ -356,20 +356,32 @@ vector<SearchParams> Setup::loadParams(
...
+  loadParams(cfg, setupFor, &paramss);
+  return paramss;
+}
+
+void Setup::loadParams(
+  ConfigParser& cfg,
+  setup_for_t setupFor,
+  vector<SearchParams>* paramss,
+  bool applyDefaultParams
+) {
```
tomtseng, line 369: see function comment in setup.h.

This new overload of `loadParams` now contains the main parameter loading logic, and the usual call `loadParams(ConfigParser& cfg, setup_for_t setupFor);` invokes this overload instead.

This overload is used for reloading the victim config in both victimplay and victimplay-gatekeeper. victimplay and victimplay-gatekeeper both need to reload the victim config, overwriting params listed in the victim config and leaving all other params untouched. Making `paramss` an input parameter instead the return value makes this overwriting possible

The diff is big but it's mainly replacing `else { }` with `else if(applyDefaultParams) { }`. Also, this overload applies its changes to the input parameter `paramss` instead of returning `paramss`.

cpp/command/selfplay.cpp
```diff
@@ -524,41 +522,27 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
 
         if(FileUtils::exists(victimCfgReloadPath)) {
           ConfigParser victimCfg;
-          std::string paramsChangeLog;
           try {
             victimCfg.initialize(victimCfgReloadPath);
             {
               lock_guard<mutex> lock(paramsReloadMutex);
```
tomtseng, line 530: here we rewrite the victim config reloading logic in `selfplay.cpp` to use `Setup::loadParams()`, reusing config reloading logic between `selfplay.cpp` and `gatekeeper.cpp`

cpp/command/gatekeeper.cpp
```diff
@@ -59,17 +61,21 @@ namespace {
     int numGamesTallied;
     double numBaselineWinPoints;
     double numCandidateWinPoints;
+    // If true, then break the game loop early if one model is going to achieve a majority of points.
+    bool terminateEarlyOnPointMajority;
```
tomtseng, line 65: setting this to `true` retains the original behavior of `gatekeeper`, but we don't want this for the victimplay gatekeeper

cpp/command/gatekeeper.cpp
```diff
@@ -214,25 +288,31 @@ int MainCmds::gatekeeper(const vector<string>& args) {
...
   string rejectedModelsDir;
+  string victimModelsDir;
   string sgfOutputDir;
   string selfplayDir;
   bool noAutoRejectOldModels;
   bool quitIfNoNetsToTest;
   try {
     KataGoCommandLine cmd("Test neural nets to see if they should be accepted for self-play training data generation.");
     cmd.addConfigFileArg("","");
+    cmd.addOverrideConfigArg();
```
tomtseng, line 299: this line allows us to use the `-override-config` command line flag — it's helpful for debugging

cpp/command/gatekeeper.cpp
```diff
@@ -328,63 +417,55 @@ int MainCmds::gatekeeper(const vector<string>& args) {
     Logger::logThreadUncaught("data write loop", &logger, dataWriteLoop);
   };
 
-  auto loadLatestNeuralNet =
```
tomtseng, line 331: this function is now split into `getLatestModelInfo`, `rejectOldTestModel`, `loadNNEvaluator` and `loadNetAndStuff` in order to reuse logic between selfplay/victimplay gatekeeping

cpp/command/gatekeeper.cpp
```diff
@@ -458,38 +539,31 @@ int MainCmds::gatekeeper(const vector<string>& args) {
   auto gameLoopProtected = [&logger,&gameLoop](int threadIdx) {
     Logger::logThreadUncaught("game loop", &logger, [&](){ gameLoop(threadIdx); });
   };
-
-  //Looping polling for new neural nets and loading them in
-  while(true) {
```
tomtseng, line 463: `NetAndStuff` evaluation is pulled out from this `while(true) {}` loop body into `evaluateNetAndStuff`

cpp/program/setup.cpp
```diff
@@ -356,20 +356,32 @@ vector<SearchParams> Setup::loadParams(
...
   vector<SearchParams> paramss;
-  int numBots = 1;
+  loadParams(cfg, setupFor, &paramss);
+  return paramss;
+}
+
+void Setup::loadParams(
+  ConfigParser& cfg,
+  setup_for_t setupFor,
+  vector<SearchParams>* paramss,
```
AdamGleave: What are the benefits of using a pointer `*` here rather than a reference `&`?

cpp/program/setup.cpp
```diff
@@ -390,277 +402,272 @@ vector<SearchParams> Setup::loadParams(
...
+    else if(applyDefaultParams)          params.lagBuffer = 0.0;
 
     if(cfg.contains("searchFactorAfterOnePass"+idxStr)) params.searchFactorAfterOnePass = cfg.getDouble("searchFactorAfterOnePass"+idxStr, 0.0, 1.0);
     else if(cfg.contains("searchFactorAfterOnePass"))   params.searchFactorAfterOnePass = cfg.getDouble("searchFactorAfterOnePass",        0.0, 1.0);
     if(cfg.contains("searchFactorAfterTwoPass"+idxStr)) params.searchFactorAfterTwoPass = cfg.getDouble("searchFactorAfterTwoPass"+idxStr, 0.0, 1.0);
     else if(cfg.contains("searchFactorAfterTwoPass"))   params.searchFactorAfterTwoPass = cfg.getDouble("searchFactorAfterTwoPass",        0.0, 1.0);
 
     if(cfg.contains("numSearchThreads"+idxStr)) params.numThreads = cfg.getInt("numSearchThreads"+idxStr, 1, 4096);
-    else                                        params.numThreads = cfg.getInt("numSearchThreads",        1, 4096);
+    else if(applyDefaultParams)                 params.numThreads = cfg.getInt("numSearchThreads",        1, 4096);
```
AdamGleave: This has the semantics of overwriting `params.numThreads` only when `numSearchThreadsN` is specified, not `numSearchThreads`. I can see a case for it: it allows `numSearchThreadsN` in the original config to continue to override the `numSearchThreads` fall-back. But it seems inconsistent with behavior elsewhere in the file. Can we either make it consistent or document the exception?

cpp/command/gatekeeper.cpp
```diff
@@ -514,64 +588,211 @@ int MainCmds::gatekeeper(const vector<string>& args) {
...
+      }
+      victimModelInfo->name = "victim-" + victimModelInfo->name;
+      string victimCfgContents;
+      ConfigParser victimCfg;
+      if(FileUtils::exists(victimCfgReloadPath)) {
+        try {
+          victimCfg.initialize(victimCfgReloadPath);
+          victimCfgContents = victimCfg.getAllKeyVals();
+        } catch (const IOError &e) {
+          logger.write(string("Victim config reloading error: ") + e.what());
```
AdamGleave, line 650: Here we seem to log but just carry on. What state is `victimCfg` in? What consequences might that have?

cpp/command/gatekeeper.cpp
```diff
@@ -514,64 +588,211 @@ int MainCmds::gatekeeper(const vector<string>& args) {
...
+        victimCfgContents,
+      };
+      if (!acceptedModelMatchup.isSameMatchup(lastAcceptedModelResults)) {
+        // We need to re-evaluate accepted model vs. victim model.
+
+        if (victimCfgContents != lastAcceptedModelResults.victimCfgContents) {
+          // Update `paramss` with the new victim config.
+          logger.write("Old victim config:\n" + lastAcceptedModelResults.victimCfgContents);
+          logger.write("Reloading with config:\n" + victimCfgContents);
+          Setup::loadParams(
```
AdamGleave, line 667: Overriding the current `paramss` introduces a path dependency: it depends not just on the default config parameters and current victim config, but all the _previous_ ones. So in theory, we could end up with a different config if the gatekeeper was killed and restarted than if it continued running.

This isn't a problem so long as the loaded `victimCfg` always contains a fixed set of keys. I think that'd usually be true with curriculum. But it might be worth explicitly documenting this assumption? Or perhaps we can just save the base config somewhere, and then copy it over to `paramss` then merge it with `victimCfg`, which should avoid path dependency. (In either case, probably want to update selfplay config reloading to be consistent with this.)

AdamGleave (commented):
Thanks for doing this! More complex than we expected going in, but a useful feature to have. Mostly looks good, left a few relatively minor comments.

cpp/program/setup.cpp
tomtseng: Thanks, I did a find/replace to make the changes in this file and didn't notice that `numThreads` is handled differently. I'll make the behavior consistent with the behavior elsewhere in the file.

cpp/command/gatekeeper.cpp
tomtseng: Looks like this `catch` was written by Viktor several months ago. I'm not sure when we would hit this error except for if the victim config is malformed. I would guess that throwing an error would be OK here and more transparent

cpp/command/gatekeeper.cpp
tomtseng: good point, I've changed it to reload the original params first before loading the victim config

cpp/program/setup.cpp
tomtseng: I'll change this to a reference. 

The rationale for a pointer: If the convention of making output parameters pointers is followed, it makes it clearer at the call site that this is a parameter that may be modified by the called function since you have call `loadParams(..., &params, ...)` instead of just `loadParams(..., params, ...)`. I'm used to this convention since I saw it in the Google C++ style guide, and it was in an internal style guide of a former employer. But it looks like Google's C++ style guide has since removed this convention.

cpp/command/selfplay.cpp
```diff
@@ -475,11 +476,13 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
     &gameRunner,
     &manager,
     &logger,
+    &cfg,
```
AdamGleave: Where is this used?

AdamGleave (approved):
LGTM. One variable I think is unused you may want to stop closing over. Only other outstanding this is the exception handler: I don't have strong feelings either way, up to you if you make it a hard error or keep it as a warning.

PR #75, opened 2022-12-29
evaluate_loop: Don't allow victim.cfg to be the victim

tomtseng:
If `victim.cfg` is the latest file in `<training-dir>/models/`, then `evaluate_loop.sh` will erroneously pick `victim.cfg` as the victim. This PR makes `evaluate_loop.sh` filter `models/` so as to not pick `victim.cfg`.


AdamGleave (approved):
LGTM

PR #76, opened 2022-12-30
gtp: Check maxVisits0 for time control

tomtseng:
I had an issue with GTP where I had set `maxVisits0` to a high number but only saw visit counts around ~4000 when I turned on `logSearchInfo = true`. 

Fix: Don't set the automatic time control if `maxVisits0` exists.

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -1684,7 +1684,13 @@ int MainCmds::gtp(const vector<string>& args) {
   engine->setOrResetBoardSize(cfg,logger,seedRand,defaultBoardXSize,defaultBoardYSize,logger.isLoggingToStderr());
 
   //If nobody specified any time limit in any way, then assume a relatively fast time control
-  if(!cfg.contains("maxPlayouts") && !cfg.contains("maxVisits") && !cfg.contains("maxTime")) {
+  if(!cfg.contains("maxPlayouts")
+      && !cfg.contains("maxPlayouts0")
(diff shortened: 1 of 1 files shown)


cpp/command/gtp.cpp
```diff
@@ -1684,7 +1684,13 @@ int MainCmds::gtp(const vector<string>& args) {
   engine->setOrResetBoardSize(cfg,logger,seedRand,defaultBoardXSize,defaultBoardYSize,logger.isLoggingToStderr());
 
   //If nobody specified any time limit in any way, then assume a relatively fast time control
-  if(!cfg.contains("maxPlayouts") && !cfg.contains("maxVisits") && !cfg.contains("maxTime")) {
+  if(!cfg.contains("maxPlayouts")
+      && !cfg.contains("maxPlayouts0")
+      && !cfg.contains("maxVisits")
+      && !cfg.contains("maxVisits0")
```
AdamGleave, line 1690: Is it safe to just checks for `max{Playouts,Visits}0` and not `max{Playouts,Visits}1`?

tomtseng: I think we don't necessarily expect it to be specified. For GTP, bot 1 is the victim and is only used for evaluating victim nodes in AMCTS. When using AMCTS-S or AMCTS-S++ you don't _need_ to specify `maxVisits1` or `maxPlayouts1`

AdamGleave (approved):
LGTM

PR #77, opened 2023-01-11
searchresults: Stricter pass hardening against seki situations

tomtseng:
Kellin noticed that hardened victims play poorly in seki situations. The proposed fix: "one can pass if only legal moves are in own pass-alive territory, in a single space one surrounds, or put oneself in atari"

## Testing

ran a match using an adversary that usually wins against cp505h but loses against cp505. the result was that the adversary lost against both cp505 and cp505h, as desired (data: `/nas/ucb/k8/go-attack/match/ttseng-warm63-vs-cp505-seki-20230110-204209`)

### DIFF
--- a/cpp/game/board.cpp
+++ b/cpp/game/board.cpp
@@ -469,7 +469,7 @@ bool Board::isLegalIgnoringKo(Loc loc, Player pla, bool isMultiStoneSuicideLegal
 }
 
 //Check if this location contains a simple eye for the specified player.
-bool Board::isSimpleEye(Loc loc, Player pla) const
+bool Board::isSimpleEye(Loc loc, Player pla, bool allowFalseEye) const
 {
   if(colors[loc] != C_EMPTY)
     return false;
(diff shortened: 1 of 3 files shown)


tomtseng:
closing this PR. The issues I encountered with this proposal:

* If I suppress all moves s.t. the placed stone will be in atari, the bot will consistently make bad blunders (see SGFs from [this training run](https://www.notion.so/chaiberkeley/adv-600-vs-cp505-v1-cp63-warmstart-lr-scale-0-05-pass-hardening-that-handles-seki-situations-9026f001dc084d0eb46de32c4f4e9163))
* If I don't suppress moves and just allow pass when the only moves are as Kellin described, you can still end up in situations where katago is in a seki position, there's some other (still bad) move elsewhere that causes passes to be suppressed, and then katago plays in a way that blunders the seki. (I didn't run a training run for this but I saw that if I loaded some SGFs with seki positions, katago would still blunder the seki rather than passing)

PR #79, opened 2023-02-01
victimplay: Print victim config reload only once, not once for each thread

tomtseng:
PR #74 updated the victimplay victim config reloading logic so that it could be reused for the gatekeeper. But I had a bug where if the config got updated, every game thread would try to load the config and print a message about the config getting updated. We only need one thread to load the config, and we only want to print the message once. (I thought I fixed this in PR #74, but maybe I forgot to push a commit.)

Changes:
* Only the first thread to notice the config change will load the config into `paramss` and print a message about the config getting updated

### DIFF
--- a/cpp/command/selfplay.cpp
+++ b/cpp/command/selfplay.cpp
@@ -485,6 +485,7 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
     &originalParamss,
     &baseParams,
     &paramsReloadMutex,
+    &lastVictimCfgContents,
     &gameSeedBase,
     &victimplay,
     &reloadVictims,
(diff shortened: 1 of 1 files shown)


AdamGleave (commented):
Is the assignment on line 544:
```
                lastVictimCfgContents = std::move(victimCfgContents);
```
guaranteed to be atomic? Otherwise I think there could be a race.

I don't understand C++ well enough to be sure either way, but it looks like `std::atomic` may help here: https://stackoverflow.com/a/8920183/14507224

tomtseng:
> Is the assignment on line 544:
> guaranteed to be atomic? Otherwise I think there could be a race.

Here we prevent races by grabbing `lock_guard<mutex> lock(paramsReloadMutex);`. I've added a comment on `paramsReloadMutex` to remind contributors to use the mutex when there's multi-threaded access to `paramss` or `lastVictimCfgContents`.

AdamGleave (approved):
LGTM. I falsely assumed `victimCfgContents` was read outside that mutex-guarded section but I see it's not so all good!

PR #80, opened 2023-02-11
search: Fix `temperatureScaleProbs()` crashing on zero temperature

tomtseng:
## Context

Running `runsearchtests.sh` hits the following assertion error:
```
katago: /engines/KataGo-custom/cpp/search/search.cpp:1041: static void Search::temperatureScaleProbs(const double*, int, double, double*, bool): Assertion `sum > 0.0' failed.
``` 

`temperatureScaleProbs()` is a function we added for logit lens. It fails in this manner when its `temperature` is zero because a division by `temperature` makes `sum` be `nan`.

## Changes

Special-case `temperatureScaleProbs()` to handle the `temperature == 0` case. The tests now run without crashing.

### DIFF
--- a/cpp/search/search.cpp
+++ b/cpp/search/search.cpp
@@ -1027,6 +1027,7 @@ void Search::temperatureScaleProbs(const double* relativeProbs, int numRelativeP
   }
   assert(maxValue > 0.0);
 
+  // Special case for zero temperature to avoid division by zero.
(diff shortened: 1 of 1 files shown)


AdamGleave (approved):
LGTM -- I suggest adding a one-liner comment to explain why we need the `if (temperature == 0.0)` block

PR #81, opened 2023-02-14
evaluate_loop.sh: Evaluate the first adversary if warmstarting

tomtseng:
If we warmstart, `evaluate_loop.sh` does not evaluate the first adversary, t0-s0-d0. This PR fixes that.


AdamGleave (approved):
LGTM

PR #82, opened 2023-02-24
search: Allow both bots to use AMCTS

tomtseng:
The assertion `assert(!oppParams.usingAdversarialAlgo())` fails when two AMCTS bots play against each other. We should allow AMCTS bots to play as long as they're not doing recursive modeling — we just want to avoid infinite recursion.

### DIFF
--- a/cpp/search/search.cpp
+++ b/cpp/search/search.cpp
@@ -504,12 +504,14 @@ Search::Search(
 {
   if (searchParams.usingAdversarialAlgo()) {
     assert(searchParams.numThreads == 1); // We do not support multithreading with AMCTS (yet).
-    assert(oppNNEval != nullptr);
+    assert(oppNNEval != nullptr || searchParams.maxVisits == 1);
 
     oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
       ? params.oppVisitsOverride.value_or(oppParams.maxVisits)
       : 1;
-    assert(!oppParams.usingAdversarialAlgo() || oppParams.maxVisits == 1);
+    // We don't want to recursively model an opponent also using A-MCTS or else
+    // we'll have infinite recursion.
+    assert(!oppParams.usingAdversarialAlgo() || searchParams.maxVisits == 1 || oppParams.maxVisits == 1);
     oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
     oppBot = make_unique<Search>(oppParams, oppNNEval, logger, rSeed + "-victim-model");
   }
--- a/cpp/search/search.cpp
+++ b/cpp/search/search.cpp
@@ -513,7 +513,8 @@ Search::Search(
     // we'll have infinite recursion.
     assert(!oppParams.usingAdversarialAlgo() || searchParams.maxVisits == 1 || oppParams.maxVisits == 1);
     oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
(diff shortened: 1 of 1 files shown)


cpp/search/search.cpp
```diff
@@ -504,14 +504,18 @@ Search::Search(
...
-      : 1;
-    oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
-    oppBot = make_unique<Search>(oppParams, oppNNEval, logger, rSeed + "-victim-model");
+    if (searchParams.maxVisits > 1) {
+      assert(oppNNEval != nullptr);
+      oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
+        ? params.oppVisitsOverride.value_or(oppParams.maxVisits)
+        : 1;
+      // We don't want to recursively model an opponent also using A-MCTS or else
+      // we'll have infinite recursion.
```
ed1d1a8d, line 513: Reminds me of https://acritch.com/osgt-is-weird/.

cpp/search/search.cpp
```diff
@@ -504,14 +504,18 @@ Search::Search(
...
-    assert(oppNNEval != nullptr);
-    assert(!oppParams.usingAdversarialAlgo());
-
-    oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
-      ? params.oppVisitsOverride.value_or(oppParams.maxVisits)
-      : 1;
-    oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
-    oppBot = make_unique<Search>(oppParams, oppNNEval, logger, rSeed + "-victim-model");
+    if (searchParams.maxVisits > 1) {
+      assert(oppNNEval != nullptr);
```
ed1d1a8d, line 508:
```suggestion
      assert(oppNNEval != nullptr);
      
```
Just adding a space afterwards to group related logic to make things easier to read.

cpp/search/search.cpp
```diff
@@ -504,14 +504,18 @@ Search::Search(
...
+      oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
+        ? params.oppVisitsOverride.value_or(oppParams.maxVisits)
+        : 1;
+      // We don't want to recursively model an opponent also using A-MCTS or else
+      // we'll have infinite recursion.
+      assert(!oppParams.usingAdversarialAlgo() || oppParams.maxVisits == 1);
+      oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
+      SearchParams oppOppParams = oppParams.usingAdversarialAlgo() ? searchParams : SearchParams();
```
ed1d1a8d: Is this necessary?

Per https://github.com/AlignmentResearch/KataGo-custom/blob/e4451f7d5bde8d1a252ac13e3ac66225d0bd242a/cpp/search/search.h#L391-L398 this line only has an effect if `oppParams.usingAdversarialAlgo()` is true. But if `oppParams.usingAdversarialAlgo()` is true, the assert above guarantees `oppParams.maxVisits == 1`, which means it doesn't matter what we set as `oppOppParams`?

ed1d1a8d (changes requested):
Short PR but surprisingly tricky logic.
Left just one comment -- I think one of the lines of logic is a noop?

cpp/search/search.cpp
tomtseng: good point, we can remove this — I had added this line before adding the surrounding `if` block that makes this unnecessary

ed1d1a8d (approved):
LGTM!

PR #83, opened 2023-04-03
selfplay: Allow selfplay games during victimplay (for iterated adversarial training)

tomtseng:
Setting up basic implementation of iterated adversarial training.
Associated `go_attack` PR: https://github.com/AlignmentResearch/go_attack/pull/106

Changes:
* `selfplay.cpp`: 
  * When performing victimplay, the config parameter `selfplayProportion` can be used to specify a proportion of games to be adversary-vs-adversary selfplay games instead of victimplay games. The purpose of this is to preserve the general competency of the adversary (e.g., if we're finetuning a KataGo model but still want it to be good at general Go play).
  * Fix victimplay workers crashing due to not finding `nnVictimPath` if they launch before curriculum creates the `nnVictimPath` directory. (With standard training is OK-ish because the victimplay workers just crash a few times at launch until the curriculum worker runs, but in iterated training the victimplay workers could potentially crash at every new iteration.)
* `curriculum.py`: Update to ignore selfplay games
* `evaluate_loop.sh`: Allow specifying different config for evaluation. (For iterated adversarial training, the config parameters should be different per iteration depending on whether we're training the cyclic-adversary against a KataGo network or training a KataGo network against the cyclic-adversary.)

### DIFF
--- a/cpp/command/selfplay.cpp
+++ b/cpp/command/selfplay.cpp
@@ -364,8 +364,10 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
       firstFileRandMinProp, dataBoardLen, dataBoardLen, onlyWriteEvery, Global::uint64ToHexString(rand.nextUInt64())
     );
 
-    tdataWriter->forVictimPlay = victimplay;
-    vdataWriter->forVictimPlay = victimplay;
+    tdataWriter->forVictimplay = victimplay;
+    vdataWriter->forVictimplay = victimplay;
+    tDataWriter->allowSelfplayInVictimplay = selfplayProportion > 0.0;
+    vDataWriter->allowSelfplayInVictimplay = selfplayProportion > 0.0;
 
     tdataWriter->useAuxPolicyTarget = cfg.getBool("useAuxPolicyTarget");
     vdataWriter->useAuxPolicyTarget = cfg.getBool("useAuxPolicyTarget");
--- a/cpp/command/selfplay.cpp
+++ b/cpp/command/selfplay.cpp
@@ -184,6 +184,13 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
   const bool switchNetsMidGame = cfg.getBool("switchNetsMidGame");
   assert(!(victimplay && switchNetsMidGame));
 
+  // Proportion of selfplay games to include during victimplay training.
+  const bool selfplayProportion =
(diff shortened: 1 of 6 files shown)


cpp/command/selfplay.cpp
```diff
@@ -258,7 +265,12 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
...
-    if(FileUtils::isDirectory(nnVictimPath)) {
+    // If victim path doesn't exist yet and isn't a gzip file, assume it's a
+    // directory that has not yet been created yet.
+    const bool isDirectory =
+      (!FileUtils::exists(nnVictimPath) && !Global::isSuffix(nnVictimPath, ".gz"))
+      || FileUtils::isDirectory(nnVictimPath);
+    if(isDirectory) {
       reloadVictims = true;
     } else {
       singleVictim.reset(loadNN("victim", nnVictimPath));
```
ed1d1a8d:
```suggestion
      // We load victims from a directory.
      // A new victim is loaded every time a new victim shows up in the directory.
      reloadVictims = true;
    } else {
      // A victim is loaded a single time from a file.
      singleVictim.reset(loadNN("victim", nnVictimPath));
```

cpp/command/selfplay.cpp
```diff
@@ -608,13 +627,36 @@ int MainCmds::selfplay(const vector<string>& args, const bool victimplay) {
...
+            gameIdx % 2 == 0 ? C_BLACK : C_WHITE,
+            gameRunner, shouldStopFunc, logger,
+            gameIdx, seed, curPredictorNNEval.get()
+          );
+        } else {  // selfplay game
+          MatchPairer::BotSpec botSpecB;
+          botSpecB.botIdx = 1;
+          botSpecB.botName = nnEval->getModelName();
+          botSpecB.nnEval = nnEval;
+          botSpecB.baseParams = curAdvSearchParams;
```
ed1d1a8d, line 646:
```suggestion
          botSpecB.baseParams = curAdvSearchParams;
          // Force selfplay games to run with MCTS.
          botSpecB.baseParams.searchAlgo = SearchParams::SearchAlgorithm::MCTS;
```

ed1d1a8d (changes requested):
Just requested a few changes. One adding some comments and the other forcing selfplay games to use MCTS.

ed1d1a8d (approved):
LGTM

PR #84, opened 2023-04-07
Reuse active-experiment in evaluate-loop

tomtseng:
motivation: i sometimes change params in active-experiment.cfg but forget to also change the param in eval

associated go_attack PR: https://github.com/AlignmentResearch/go_attack/pull/110


tomtseng:
I haven't thought much about this issue in a while so I'm just going to close this

PR #85, opened 2023-05-03
gtp: Don't load victim model if using MCTS

tomtseng:
Issue:
* GTP has both the `-model` and `-victim-model` flag, with the `-victim-model` used to specify the victim for A-MCTS. If the user provides the `-victim-model` flag but is using MCTS, the victim model is still loaded despite being unused, taking up GPU memory.

Fix:
* Don't load the victim model if MCTS is used.

Alternatively, we could throw an error if the user sets the `-victim-model` flag when using MCTS. But this would make our launch script for transfer experiments more annoying, since the launch script always sets the `-victim-model` (it would either need to parse the KataGo config or be told by the user to determine whether MCTS/AMCTS is being used).

### DIFF
--- a/cpp/command/gtp.cpp
+++ b/cpp/command/gtp.cpp
@@ -469,7 +469,17 @@ struct GTPEngine {
 
     nnEval = getNNEvaluator(nnModelFile, nnModelFile);
     if (opModelFile != "") {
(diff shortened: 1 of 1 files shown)


ed1d1a8d (approved):
LGTM 👍

PR #87, opened 2023-05-10
Update Search constructor with better comments

ed1d1a8d:
In particular, update a comment to include a better reason to do an assert.

Also update the structure of the code and format to make it easier to read.

Logic should be unchanged.


ed1d1a8d:
Actually closed because this PR is better left as a comment in Tom's PR.

PR #88, opened 2023-05-12
getAnalysisJson: Handle nullptr

tomtseng:
Issue: When running  `./runoutputtests.sh`, `getAnalysisJson()` crashes on the test `Testing 1000 visits just before terminal position, then playing the pass and having tree reuse. allowDirectPolicyMoves false...` due to hitting a `nullptr`. (We stopped handling `nullptr` correctly when we refactored `getAnalysisJson()` to support `includeTree`.)

Fix: return empty JSON if `getAnalysisJson()` is called on a `nullptr` node

### DIFF
--- a/cpp/search/searchresults.cpp
+++ b/cpp/search/searchresults.cpp
@@ -1869,6 +1869,9 @@ bool Search::getAnalysisJson(
     static constexpr int minMoves = 0;
 
     json moveInfos = json::array();
+    if(node == NULL) {
+      return moveInfos;
+    }
     Player pla = node->nextPla;
 
     vector<AnalysisData> buf;


ed1d1a8d:
Hmm, I'm running the old code (without the null check) and `./runoutputtests.sh` seems to run fine. Am I missing something?

tomtseng:
What I'm seeing without the null check is that we reach `Testing 1000 visits just before terminal position, then playing the pass and having tree reuse. allowDirectPolicyMoves false...`, and then we stop before outputting `getAnalysisJson success: 0`. This is different from the expected output existing in `runOutputTests.txt` where after `Testing 1000 visits...` we see further tests like `Running training write tests`

ed1d1a8d:
Ah gotchu, this make sense now. Fix looks good to me.

ed1d1a8d (approved):

PR #89, opened 2023-05-14
Merge upstream v1.12.0

tomtseng:
(This PR should not be squashed.)

Biggest merge conflict is that `train.py` has been rewritten to use PyTorch instead of TF, so most of our `train.py` changes got discarded except for the `disable-vtimeloss` flag.
The switch to PyTorch causes another issue: we can't warmstart from models trained with KataGo v1.11 or earlier since the training weight format has changed.

Changes we'll need to make to `go_attack` (implemented in https://github.com/AlignmentResearch/go_attack/pull/118):
* Add `torch` and `requests_toolbelt` to `compose/python/requirements.txt` so that the Python image used by our linting CircleCI job can `pytype` the `KataGo-custom` Python code without complaining about missing libraries
* At least for now, we'll still want to keep the TF training code around for warmstarting on old models — either keep a old branch of KataGo in `engines/` or copy the old TF training code into a separate directory `KataGo-custom` 
* If we want to finetune the new architectures we'll need to edit `kubernetes/train.sh` since some of the flags to `train.py` have changed (e.g., model kind is a flag `-model-kind` instead of a positional argument, and warmstarting is done with the `-initial-checkpoint` flag instead of done by copying files into a `initial_weights` directory)


cpp/tests/testamcts.cpp
```diff
@@ -555,21 +555,23 @@ void AMCTSTests::checkFinalMoveSelection(const Search& bot) {
       double fpuValue = bot.getFpuValueForChildrenAssumeVisited(
           *tree.root, tree.root->nextPla, true, 1.0, parentUtility,
           parentWeightPerVisit, parentUtilityStdevFactor);
+      double exploreScaling = bot.getExploreScaling(totalChildWeight, parentUtilityStdevFactor);
```
ed1d1a8d:
```suggestion
      const double exploreScaling = bot.getExploreScaling(totalChildWeight, parentUtilityStdevFactor);
```

cpp/tests/testamcts.cpp
```diff
@@ -555,21 +555,23 @@ void AMCTSTests::checkFinalMoveSelection(const Search& bot) {
...
       return bot.getExploreSelectionValueOfChild(
           *tree.root, policyProbs, heaviestChild,
-          tree.revEdge.at(heaviestChild).moveLoc, totalChildWeight,
+          tree.revEdge.at(heaviestChild).moveLoc, exploreScaling, totalChildWeight,
           heaviestChild->stats.visits.load(std::memory_order_acquire), fpuValue,
-          parentUtility, parentWeightPerVisit, parentUtilityStdevFactor, false,
+          parentUtility, parentWeightPerVisit, false,
           false, maxChildWeight, NULL);
     }();
+    double exploreScaling = bot.getExploreScaling(totalChildWeight, 1.0);
```
ed1d1a8d:
```suggestion
    const double exploreScaling = bot.getExploreScaling(totalChildWeight, 1.0);
```

ed1d1a8d (changes requested):
Have checked over all non python/* files and those look good.

For the python/* files, since the pytorch code seems to be non backwards compatible with the tensorflow code (per lightvector's checkpoint comments), I'm thinking we just copy over the python/* files exactly as is from the v1.12.0 code. If we need to finetune an older network, we can just use an older version of the repo.

tomtseng:
(Context to make this conversation clearer to others: We now cannot warmstart from old <=v1.11 models since they're saved as TF checkpoints whereas v1.12.0 has switched to PyTorch and expects PyTorch checkpoints. I asked on Discord whether there was a nice way to convert from TF checkpoints to PyTorch checkpoints and lightvector said there's no official way. Someone else suggested trying [MMdnn](https://github.com/Microsoft/MMdnn) for conversion)

The Python files were essentially just copied over except for `-disable-vtimeloss` plus some changes to make the Python linter happy. If we don't think we'll care about `-disable-vtimeloss` for finetuning the newer KataGo architectures then yeah, let's just copy the v1.12.0 Python code without changes and disable linting for those files. 
I think removing `-disable-vtimeloss` is fine — the newer KataGo architectures were all trained with vtimeloss and it seems consistent to finetune them with vtimeloss as well. But we used `-disable-vtimeloss` for training our adversaries, and we might care about that eventually (e.g., if we work on adding vision transformers and think it's easier to write them in PyTorch instead of TF).

It does seem like converting checkpoints from TF to PyTorch will take some effort and it's risky to devote time to it right now. 
I think I'll add another copy of KataGo in `/go_attack/engines` that has the old TF training code, since we'll be using it frequently. In particular, whenever we want to finetune our cyclic adversary, we're using its TF checkpoint. I'll spend today setting up a finetuning run of the cyclic adversary vs. a b18c384nbt net and see if anything breaks.

ed1d1a8d:
I'd actually vote the leave out vtimeloss completely. We can make a commit to add it back in once we start running experiments that actually require disabling vtimeloss. I'm afraid of adding it back in right now without proper testing.

As for how to finetune our old models, I'd vote for your suggestion of copying the old TF training code into a separate directory KataGo-custom.

tomtseng:
OK, I removed the `-disable-vtimeloss` flag and made a draft PR in case we want to add the flag back in: https://github.com/AlignmentResearch/KataGo-custom/pull/98

In the corresponding `go_attack` PR https://github.com/AlignmentResearch/go_attack/pull/118 I made the old TF training code a separate submodule. I think I like this a little more because it preserves the Git history of the TF training files (easier to Git Blame) 

EDIT Nov 17: Oops, after that PR #98 got approved, I merged it without remembering that the target branch was this branch. So now the `-disable-vtimeloss` is added back in, but it's been tested (albeit not with a full training run)

ed1d1a8d (approved):
Basically lgtm.

One clarification question -- I saw that `python/train.py` had number of changes compared to upstream. I'm guessing most of these were done to make linting pass? If so then this is good to merge IMO. If not then maybe we should revert the changes to make the file to match upstream more...

PR #93, opened 2023-05-15
Make codespell ignore model.py

ed1d1a8d:
[Previously](https://app.circleci.com/jobs/github/AlignmentResearch/KataGo-custom/1469) was failing with the error:
```
./python/model.py:386: pythong ==> python
```
We fix this by ignoring `./python/model.py`

Not sure why why codespell started failing all of a sudden...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,7 @@ jobs:
 
       - run:
           name: codespell
-          command: codespell -I .codespell.skip --skip='*.pyc,*.ipynb' ./python/
+          command: codespell -I .codespell.skip --skip='*.pyc,*.ipynb,python/model.py' ./python/
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,7 +135,7 @@ jobs:
 
       - run:
           name: codespell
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -135,7 +135,7 @@ jobs:
 
       - run:
           name: codespell
-          command: codespell -I .codespell.skip --skip='*.pyc,*.ipynb' ./python/
+          command: codespell -I .codespell.skip --skip='*.pyc,*.ipynb,python/model.py' ./python/
```
ed1d1a8d:
```suggestion
          command: codespell -I .codespell.skip --skip='*.pyc,*.ipynb,./python/model.py' ./python/
```

tomtseng (approved):
Weird, I wonder if it's because I pushed a new `:python-deps` Docker image, which our CircleCI job uses. But I think the only thing I did was install `torch` and `requests_toolbelt` so that the PyTorch code in v1.12 would pass `pytype` — I don't see why this would cause `codespell` to fail on `stable`.

Anyway, seems fine, though I feel I would just fix the typo in `model.py` instead of ignoring `model.py`
(EDIT: Actually v1.12.0 just switches to `model_pytorch.py` and `model.py` is unused, so we'll want to end up ignoring `model.py` anyway)

PR #95, opened 2023-05-25
evaluate_loop: Fix victim selection

tomtseng:
Two bugs with `evaluate_loop.sh`:
1. If I specify multiple victims with the flag `--victim-list`, we expect each victim to be evaluated on each new adversary. However, only the first victim gets evaluated.
    - Why: After launching one match with an adversary at a particular step count `$STEP`, we update `LAST_STEP="$STEP"` and don't evaluate any adversaries with step count `<= $LAST_STEP`. But this update occurs within the `for` loop over victims, so all victims after the first victim don't launch a match since the adversary has step count `== $LAST_STEP`
2. If I don't specify `--victim-list`, we expect the victim to be the latest victim at the time each new adversary appears. However, the victim does not get updated when a new victim appears
    - Why: The `--victim-list` argument is assigned to variable `VICTIM_LIST`. We check for `--victim-list` not being set by checking `VICTIM_LIST` is empty. If `VICTIM_LIST` is empty, then we look for the latest victim and assign it to `VICTIM_LIST`. But then on the next iteration, `VICTIM_LIST` is non-empty, so it's never updated to re-look for the latest victim.

Fixes:
1. Move the `LAST_STEP` update outside the `for` loop over victims.
2. Assign `--victim-list` to a different argument `VICTIM_LIST_ARG` and check this arg to determine whether to look for the latest victim for assignment to `VICTIM_LIST`

### DIFF
--- a/cpp/evaluate_loop.sh
+++ b/cpp/evaluate_loop.sh
@@ -33,7 +33,7 @@ usage() {
     echo "Currently expects to be run from within the 'cpp' directory of the KataGo repo."
 }
 
-VICTIM_LIST=""
+VICTIM_LIST_ARG=""
 VICTIMS_DIR=""
 PREDICTOR_DIR=""
 KATAGO_BIN="/engines/KataGo-custom/cpp/katago"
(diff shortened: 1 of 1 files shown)


UFO-101:
LGTM. Maybe you could come up with a more descriptive name than `VICTIM_LIST_ARG`?

UFO-101 (approved):

PR #98, opened 2023-08-23
train.py: Add -disable-vtimeloss to PyTorch training

tomtseng:
With the old TF training code we had a flag `-disable-vtimeloss` to disable the variance time loss. We removed it when we moved to KataGo v1.12.0, which changes the training code to PyTorch, since we didn't want to add extraneous changes to the PyTorch code and we weren't using the PyTorch code.

This PR adds the `-disable-vtimeloss` flag back.

Testing: Checked that running `train.py` with `-disable-vtimeloss` drives the loss down without driving vtimeloss down.

### DIFF
--- a/python/train.py
+++ b/python/train.py
@@ -91,7 +91,7 @@ if __name__ == "__main__":
   parser.add_argument('-intermediate-loss-scale', type=float, help='Loss factor scale for intermediate head', required=False)
   parser.add_argument('-intermediate-distill-scale', type=float, help='Distill factor scale for intermediate head', required=False)
 
-  parser.add_argument('-disable-vtimeloss', help='Disable vtimeloss for training', required=False, action='store_true')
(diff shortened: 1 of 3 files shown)


python/train.py
```diff
@@ -91,6 +91,8 @@
   parser.add_argument('-intermediate-loss-scale', type=float, help='Loss factor scale for intermediate head', required=False)
   parser.add_argument('-intermediate-distill-scale', type=float, help='Distill factor scale for intermediate head', required=False)
 
+  parser.add_argument('-disable-vtimeloss', help='Disable vtimeloss for training', required=False, action='store_true')
```
AdamGleave: Alternative:
```suggestion
  parser.add_argument('-disable-vtimeloss', help='Disable vtimeloss for training', required=False, dest='use_vtimeloss', action='store_false')
```
then don't need the `not` later.

Not sure this is actually any better though but figured I'd give the option as the `dest` argument is easily overlooked.

AdamGleave (approved):
LGTM

PR #99, opened 2023-10-10
Search: Fix segfault by initializing oppBot for A-MCTS

tomtseng:
### Bug 
Say I'm running `victimplay` with one of the bots running A-MCTS with one visit. Typically in A-MCTS, the bot will initialize `oppBot` to model the opponent bot. However, we don't initialize `oppBot` with the number of visits is 1 because then the oppBot will never be evaluated during the course of the game. This check on the number of visits prevents two A-MCTS bots from modeling each other in an infinite loop (i.e., A-MCTS bot 0 initializes oppBot with 1 visit to model A-MCTS bot 1, and then oppBot->oppBot is not initialized due to the check on visits == 1).

But `victimplay` actually runs the bots with a different number of visits at the end of the game in order to, e.g., compute the lead. Then `oppBot` being uninitialized causes a segfault.

### Fix

Initialize `oppBot` based on whether the argument `oppNNEval` is provided rather than based on the number of visits. Then a 1-visit A-MCTS bot will initialize oppBot, but oppBot->oppBot is still not initialized as desired.

### Testing

Run victimplay with both bots using 1 visit and AMCTS , check it doesn't crash:
`./katago victimplay -config /go_attack/configs/active-experiment.cfg -config /go_attack/configs/compute/1gpu.cfg   -config /go_attack/configs/extra.cfg -models-dir /nas/ucb/k8/go-attack/victimplay/ttseng-gating-on-20221229-151219/models -nn-victim-path /nas/ucb/k8/go-attack/victimplay/ttseng-gating-on-20221229-151219/victims  -max-games-total 4 -output-dir /tmp/asdf`
with `extra.cfg` being:
```
bSizes=5
bSizeRelProbs=1
komiByBSize=6.5
logSearchInfo=true
logToStdout=true
maxVisits0=1
maxVisits1=1
searchAlgorithm1=AMCTS
useGraphSearch=false
```

### DIFF
--- a/cpp/search/search.cpp
+++ b/cpp/search/search.cpp
@@ -117,8 +117,6 @@ Search::Search(
     // When using AMCTS, we don't support graph search for ourselves or our
     // opponent.
     assert(!searchParams.useGraphSearch);
-    // We only want to initialize oppBot if maxVisits > 1 or else we'll get
-    // infinite recursion when both bots are using A-MCTS.
     if (searchParams.maxVisits > 1) {
       assert(oppNNEval != nullptr);
       oppParams.maxVisits = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_R
@@ -179,15 +177,9 @@ string Search::getRankStr() const {
            std::to_string(bot->searchParams.rootNumSymmetriesToSample);
   };
 
-  if (searchParams.usingAdversarialAlgo()) {
-    string rankStr = "algo=" + searchParams.getSearchAlgoAsStr() + "," + getVisitStr(this, "");
-    // oppBot may be null if the number of visits is 1.
-    if (oppBot != nullptr) {
-      rankStr += "," + getVisitStr(oppBot.get(), "opp_");
-    }
-    return rankStr;
-  }
-  
+  if (searchParams.usingAdversarialAlgo())
+    return "algo=" + searchParams.getSearchAlgoAsStr() + "," + getVisitStr(this, "") + "," + getVisitStr(oppBot.get(), "opp_");
+
   if (searchParams.searchAlgo == SearchParams::SearchAlgorithm::MCTS)
     return "algo=MCTS," + getVisitStr(this, "");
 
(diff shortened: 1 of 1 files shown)


cpp/search/search.cpp
```diff
@@ -114,11 +114,31 @@ Search::Search(
...
+        : 1;
+      // We don't want to recursively model an opponent also using A-MCTS or else
+      // we'll have infinite recursion.
+      assert(!oppParams.usingAdversarialAlgo() || oppParams.maxVisits == 1);
+      oppParams.rootNumSymmetriesToSample = params.searchAlgo == SearchParams::SearchAlgorithm::AMCTS_S ? 1 : oppParams.rootNumSymmetriesToSample;
+      oppBot = make_unique<Search>(oppParams, oppNNEval, logger, rSeed + "-victim-model");
+    }
+
+
     if (searchParams.maxVisits > 1) {
```
AdamGleave: Do we still need this existing code? When would we have searchParams.maxVisits > 1 but oppNNEval == nullptr? If we need both, can we reduce code duplication (body of if blocks seem identical)?

cpp/search/search.cpp
```diff
@@ -179,15 +199,9 @@ string Search::getRankStr() const {
...
-    string rankStr = "algo=" + searchParams.getSearchAlgoAsStr() + "," + getVisitStr(this, "");
-    // oppBot may be null if the number of visits is 1.
-    if (oppBot != nullptr) {
-      rankStr += "," + getVisitStr(oppBot.get(), "opp_");
-    }
-    return rankStr;
-  }
-  
+  if (searchParams.usingAdversarialAlgo())
+    return "algo=" + searchParams.getSearchAlgoAsStr() + "," + getVisitStr(this, "") + "," + getVisitStr(oppBot.get(), "opp_");
```
AdamGleave, line 193: How do we know `oppBot` is always non-null? If it's always initialized what's the point of the if checks above?

cpp/search/search.cpp
tomtseng: thanks, this code should be removed — had kept it around as reference while resolving a rebase conflict but then forgot to delete it

cpp/search/search.cpp
tomtseng: For a top-level A-MCTS `Search` struct that users initialize, the checks above should make it so that `oppBot` is always be initialized, whereas `oppBot->oppBot` is not. But I guess it would be coherent to call this function on an `oppBot`, which will lead to a `nullptr` dereference, even though we never do it currently, so I'll re-add a `nullptr` check here

AdamGleave (approved):
LGTM

PR #100, opened 2023-10-19
Add ViT PyTorch model, and allow C++ code to run PyTorch models via TorchScript

tomtseng:
Motivation: Want to experiment with non-conv architectures, and in particular ViTs, without having to write C++ backend code for specially handling each of those architectures.

Changes:
* Python:
  * Add ViT architecture to `model_pytorch.py` and `modelconfigs.py`
  * Add script `export_model_as_torchscript.py` to convert PyTorch model to TorchScript model
* C++:
  * Add `pytorchbackend.cpp` that can run TorchScript models  

Associated go_attack PR: https://github.com/AlignmentResearch/go_attack/pull/126


cpp/neuralnet/cudabackend.cpp
```diff
@@ -9,6 +9,9 @@
 #include "../neuralnet/nninputs.h"
 #include "../neuralnet/nneval.h"
 #include "../neuralnet/desc.h"
+// TODO(tomtseng): We're crudely bolting the PyTorch backend here while we
+// prototype the PyTorch backend implementation, but it doesn't belong here.
+#include "../neuralnet/pytorchbackend.h"
```
ed1d1a8d: Where does it belong?

python/model_pytorch.py
```diff
@@ -1581,60 +1619,89 @@ def add_brenorm_clippage(self, upper_rclippage, lower_rclippage, dclippage):
...
-        x_spatial = self.conv_spatial(input_spatial)
-        x_global = self.linear_global(input_global).unsqueeze(-1).unsqueeze(-1)
-        out = x_spatial + x_global
-        # print("TENSOR BEFORE TRUNK")
-        # print(out)
+        if self.is_vit:
+            # Concatenate global features onto spatial features.
+            out = torch.cat([
+                input_spatial,
+                input_global.unsqueeze(-1).unsqueeze(-1).expand(-1, self.global_input_shape[0], self.pos_len, self.pos_len)
```
ed1d1a8d: Could we rewrite this in einops to make it more readable?

cpp/neuralnet/cudabackend.cpp
tomtseng: Changed this comment to be less aggressive because realistically we're not going to spend time on refactoring this:
```cpp
// note(tomtseng): We're crudely bolting the PyTorch backend here while we
// prototype the PyTorch backend implementation, but ideally we'd refactor
// things so that the PyTorch backend is independent of the other backends.
```

python/model_pytorch.py
```diff
@@ -338,10 +340,10 @@ def forward(self, x, mask, mask_sum_hw):
         """
         mask_sum_hw_sqrt_offset = torch.sqrt(mask_sum_hw) - 14.0
 
-        layer_mean = torch.sum(x, dim=(2, 3), keepdim=True, dtype=torch.float32) / mask_sum_hw
+        layer_mean = torch.sum(x, dim=(2, 3), keepdim=True) / mask_sum_hw
         # All activation functions we use right now are always greater than -1.0, and map 0 -> 0.
         # So off-board areas will equal 0, and then this max is mask-safe if we assign -1.0 to off-board areas.
-        (layer_max,_argmax) = torch.max((x+(mask-1.0)).view(x.shape[0],x.shape[1],-1).to(torch.float32), dim=2)
+        (layer_max,_argmax) = torch.max((x+(mask-1.0)).view(x.shape[0],x.shape[1],-1), dim=2)
```
ed1d1a8d, line 346: Why do we not cast to float32 anymore? Does it break torchscript fp16 tracing? Also why did we use to have this force cast in the first place?

python/model_pytorch.py
```diff
@@ -20,7 +22,7 @@ def act(activation, inplace=False):
     if activation == "mish":
         return torch.nn.Mish(inplace=inplace)
     if activation == "gelu":
-        return torch.nn.GELU(inplace=inplace)
+        return torch.nn.GELU()
```
ed1d1a8d, line 25: Why no more inplace GELU?

python/model_pytorch.py
```diff
@@ -1581,59 +1637,88 @@ def add_brenorm_clippage(self, upper_rclippage, lower_rclippage, dclippage):
...
+            # Apply mask to global features.
+            out[:,-self.global_input_shape[0]:] *= input_spatial[:, 0:1, :, :]
+
+            out = torch.nn.functional.pad(out, (0, self.padding, 0, self.padding))
+            # Output is batch_size x (sequence_length = num_patches + 1) x
+            # hidden_size.
+            out = self.vit(out).last_hidden_state
+            # Remove [CLS] patch usually used for attaching a classification
+            # head.
+            out = out[:, 1:]
```
ed1d1a8d:
```suggestion
            # Remove [CLS] patch usually used for attaching a classification
            # head. The [CLS] patch is the first patch, see
            # https://github.com/huggingface/transformers/blob/092f1fdaa4224fdd88c616dc9678e6fcb37bfffd/src/transformers/models/vit/modeling_vit.py#L131-L133
            out = out[:, 1:]
```

python/model_pytorch.py
```diff
@@ -1581,59 +1637,88 @@ def add_brenorm_clippage(self, upper_rclippage, lower_rclippage, dclippage):
...
+            out = self.unembedder(out)
+            out = einops.rearrange(
+                    out,
+                    "b (h w) (c ph pw) -> b c (h ph) (w pw)",
+                    h=self.num_patches_one_dim,
+                    w=self.num_patches_one_dim,
+                    c=self.c_trunk,
+                    ph=self.patch_size,
+                    pw=self.patch_size,
+            )
```
ed1d1a8d, line 1672: Checks out from some experimentation I did here: https://colab.research.google.com/drive/1NJPzJ7aX67U1hcAQiTOz1EEKC-p_cFvP?authuser=0#scrollTo=-qweZ-r2vkbP

ed1d1a8d (approved):
LGTM as far as I can tell. Just had one suggestion for an additional comment.
Also wrote up my account of the ViT architecture in the overleaf, probably good to check that what I wrote matches your expectation before merging.

python/model_pytorch.py
tomtseng: Yeah, this would break FP16, but it's not clear to me why this existed in the first place

python/model_pytorch.py
tomtseng: it turns out to not be a valid input to torch.nn.GELU(), maybe this codepath was never tested
https://pytorch.org/docs/stable/generated/torch.nn.GELU.html

PR #101, opened 2023-11-09
curriculum: Add option to harden victims below some visit threshold

tomtseng:
In the continuous training run I added a hack to harden the victim when visits were low. This PR turns the hack into a command-line option to `curriculum.py`.

### DIFF
--- a/python/curriculum.py
+++ b/python/curriculum.py
@@ -394,7 +394,8 @@ class Curriculum:
         config: Sequence[Config],
         stats_rolling_window: int,
         game_buffer_capacity: int,
-        harden_below_visits: Optional[int],
+        # Defaults to 0 (no victims are hardened)
+        harden_below_visits: int = 0,
     ):
         """Initial curriculum setup.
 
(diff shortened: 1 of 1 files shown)


python/curriculum.py
```diff
@@ -394,6 +394,7 @@ def __init__(
         config: Sequence[Config],
         stats_rolling_window: int,
         game_buffer_capacity: int,
+        harden_below_visits: Optional[int],
```
AdamGleave: What would happen if we did:
```suggestion
        harden_below_visits: int = 0,
```

This would avoid needing to explicitly check for None on comparisons like on line 529, but since visits always >1 the comparison would still return True if `harden_below_visits` not explicitly specified. Would this have any negative effects?

(If you do use this change would recommend adding brief explanation to docstring as it may be counterintuitive why it's 0 by default.)

python/curriculum.py
```diff
@@ -745,6 +754,15 @@ def parse_args() -> argparse.Namespace:
         default="configs/curriculum_conf.json",
         help="Curriculum JSON config with " "victims sequence (JSON file path)",
     )
+    # From a UI standpoint it makes more sense to be able to specify in the JSON
```
AdamGleave, line 755: Yeah, seems fine, we can always make it more generic later on if a use case arises. Thanks for including the explanation in the comment!

AdamGleave (approved):
LGTM -- 1 suggestion that might simplify code a bit.

PR #102, opened 2023-11-10
evaluate_loop: Add flag for adjusting the number of games

tomtseng:
`evaluate_loop.sh` is hardcoded to always run 100 games per iteration. This PR makes the number of games configurable.

Example use case: When training cp505-h* in our iterated training, I want to evaluate the latest cp505-h*-s* against the latest cyclic-r* but also against the original cyclic adversary and against cp505. Since I've added additional opponent bots (which can be done by adjusting config params), I'd also like to increase the number of games, but that's not possible without this PR.


AdamGleave (approved):
LGTM

PR #103, opened 2023-11-29
train: Add TensorBoard logging

tomtseng:
Changes:
* Add tensorboard logging back into the PyTorch training code

### DIFF
--- a/python/train.py
+++ b/python/train.py
@@ -24,6 +24,7 @@ import torch.nn
 import torch.optim
 import torch.distributed
 import torch.multiprocessing
+import torch.utils.tensorboard
 from torch.nn.parallel import DistributedDataParallel
 from torch.optim.swa_utils import AveragedModel
 from torch.cuda.amp import GradScaler, autocast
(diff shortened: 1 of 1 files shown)


python/train.py
```diff
@@ -817,12 +818,20 @@ def log_metrics(metric_sums, metric_weights, metrics, metrics_out):
       metrics_out.write(json.dumps(metrics_to_print) + "\n")
       metrics_out.flush()
 
+    num_samples = metrics_to_print.get("nsamp_train") or metrics_to_print["nsamp"]
+    for k, v in metrics_to_print.items():
```
AdamGleave: This will be logging `nsamp_train` and `nsamp` as well -- do we want to log num_samples vs num_samples?

python/train.py
```diff
@@ -1189,6 +1198,8 @@ def log_metrics(metric_sums, metric_weights, metrics, metrics_out):
 
   train_metrics_out.close()
```
AdamGleave, line 1202: Ideally this would live in a `try`, `finally:` block -- but if it's run at the end of the program probably doesn't matter too much either way.

AdamGleave (approved):
LGTM, a couple of minor comments

python/train.py
tomtseng: yeah i thought about that and didn't deal with it because i figured it didn't matter that there'd be one extra useless plot, but I'll change this code to remove the plot

PR #105, opened 2024-02-08
Add flag for scaling move limit with board size

tomtseng:
In our training run attacking b18, I saw that some small-board games were very long, e.g., an 11x11 game hit the move limit of 1600. 1600 moves is already a very generous move limit for 19x19 games. Probably we're hitting the move limit 
due to pass-alive hardening.

Let's reduce the move limit for small-board games — this PR adds a boolean config parameter `scaleMaxMovesWithBoardSize`. If enabled, the move limit is scaled by `(board area) / 19^2`.


ed1d1a8d (approved):
LGTM!

PR #108, opened 2024-03-07
pytorchbackend: Fix ViT multi-gpu support

tomtseng:
When running the C++ TorchScript back-end with multiple server threads per model, which happens when we are using multiple GPUs, we get JiT errors like
```
terminate called after throwing an instance of 'c10::Error'
  what():  method '__torch__.torch.nn.modules.linear.___torch_mangle_14.Linear.forward' already defined.
Exception raised from register_function at ../torch/csrc/jit/api/compilation_unit.h:309 (most recent call first):
```
This happens when threads concurrently try to copy a model.

This PR fixes this issue by putting a mutex around model copying.

### DIFF
--- a/cpp/neuralnet/pytorchbackend.cpp
+++ b/cpp/neuralnet/pytorchbackend.cpp
@@ -42,8 +42,12 @@ LoadedModel::LoadedModel(const std::string& filename)
   , modelName(filename) {}
 
 LoadedModel::LoadedModel(const LoadedModel& other)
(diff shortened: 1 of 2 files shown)


AdamGleave (approved):
LGTM.

Only reviewed this specific change -- I would not be surprised if there were other thread safety violations we've unknowingly introduced elsewhere in the PyTorch code :/ but probably OK if this only affects inference not training.

PR #109, opened 2024-03-12
GTP: Add config param for forcing a rule set

tomtseng:
Our bots are only trained with Tromp-Taylor rules, but KGS doesn't have Tromp-Taylor rules. If a bot plays with non-Tromp-Taylor rules, then because the rules are features of the NN input, all NN inputs will be out-of-distribution. We therefore want to force our bots to always think they're playing with Tromp-Taylor rules.

This PR adds a config param `ignoreGTPRuleChanges` (in the same vein as the existing KataGo-raw config param `ignoreGTPAndForceKomi`) that makes KataGo ignore rule changes from `katago-set-rule`, `katago-set-rules`, `kgs-rules`, and `loadsgf`.


AdamGleave (approved):
LGTM.

Bit annoying we have to patch this in three different places, but given it is three different commands not an unreasonable way to do it. I guess we could push it into `setRulesNotIncludingKomi` but then we'd have to pass in the config parameter anyway, not clearly better.

PR #112, opened 2024-04-01
cpp: Add flags for penalizing hit-move-limit games

tomtseng:
For our gift-adversary, we needed these flags to penalize games that hit the move limit.

Added config flags:
* bot setting `noResultUtility`: Sets utility of no-result games. (The `noResultUtilityForWhite` flag already exists, which makes sense for self-play where the same trained bot is playing both colors and the utility should be zero-sum to keep training consistent. It doesn't fit the situation where we want to set the utility of no-result games for the adversary in victimplay regardless of the adversary's color.)
* game setting `hitTurnLimitIsNoResult`: Mark hit-turn limit games as no-result in training, as opposed to scoring and marking win/loss based on the 
* bot setting `forceAllowNoResultPredictions`: Force no-result logits to be not zeroed. No-result games are supposed to only occur in certain rule sets, and no-result value logits are cleared out in other rule sets. This is no longer true now that `hitTurnLimitIsNoResult` exists. (I thought about not having this flag and just not clearing logits if `hitTurnLimitIsNoResult` is true, but actually we want to be able to set this on a per-bot basis — we still want the victim to be clearing its logits even if the adv is not)

### DIFF
--- a/cpp/neuralnet/nneval.cpp
+++ b/cpp/neuralnet/nneval.cpp
@@ -951,7 +951,10 @@ void NNEvaluator::evaluate(
         double shorttermWinlossErrorPreSoftplus = buf.result->shorttermWinlossError;
         double shorttermScoreErrorPreSoftplus = buf.result->shorttermScoreError;
 
-        if(history.rules.koRule != Rules::KO_SIMPLE && history.rules.scoringRule != Rules::SCORING_TERRITORY)
+        const bool shouldClearNoResult = history.rules.koRule != Rules::KO_SIMPLE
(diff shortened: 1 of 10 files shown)


AdamGleave (approved):
LGTM.

Two questions (sorry, couldn't manage to leave comments on file due to unreliable in-flight WiFi):
  - `nninputs.cpp:55`: how were these hash values generated -- is it just random? I did not verify that these were sensible in any way.
  - `searchhelpers.cpp` lines 226 and 234: I take it utility is always from white's perspective so, in order for it to be not-zero-sum, we need to flip the sign if it's actually black playing?

PR #114, opened 2024-06-01
selfplay: Make `useAuxPolicyTarget` parameter optional

tomtseng:
We added a config flag `useAuxPolicyTarget` in https://github.com/AlignmentResearch/KataGo-custom/pull/8/commits/f2ee97a99eb80d304c81bbd8b230c2b5965be376. Currently we throw an error if the param is not specified

This PR makes it so that this param is optional, so that selfplay runs can directly use configs from `engines/KataGo-raw/cpp/configs`.


AdamGleave (approved):
LGTM.

PR #116, opened 2024-11-27
pytorchbackend: Note how to run GTP on small board sizes

tomtseng:
If I `loadsgf` a 11x11 game in GTP mode with the Pytorch backend I get the error "Board len not yet supported", but it turns out there's an existing config parameter `gtpForceMaxNNSize` that can avoid this error. This PR notes in the code comment above the error that the user should set this config param.

Also modify .circleci to make GPU tests work again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,8 +19,7 @@ executors:
     environment:
       NUM_CPUS: 8
   cpp-gpu:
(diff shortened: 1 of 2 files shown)


AdamGleave (approved):
LGTM.

Do we want to include this config parameter in any of the example GTP configs?

